PR #8, opened 2021-11-02
adding devcontainer

ngam:
- .devcontainer/ has json details of settings and extensions as well as dockerfile
- linting and minor edits of readme
- added pandas to requirements


Gorkowski:
simple and clean, I do prefer the one sentence per line for text but either is fine with me :)

PR #24, opened 2021-11-18
Created particle class & methods, typing & units.

wkchuang:
__init__.py added so only 1 units registry is created.
Added a few unit tests to check for units compatibility and for positive and negative cases.
Added live share to devcontainer

### DIFF
--- a/aerosol_dynamics/__init__.py
+++ b/aerosol_dynamics/__init__.py
@@ -1,0 +1,2 @@
+from pint import UnitRegistry
+ureg = UnitRegistry()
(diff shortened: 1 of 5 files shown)


ngam:
@wkchuang did you intentionally close this? You could make it a draft or put a WIP or something... 

I skimmed it and it looks good, thanks for looking into the units!

ngam:
About rebasing/squashing --- there is an option to squash upon closing a pr, so no need to worry about squashing things ahead of time.

wkchuang:
yeah, I intentionally closed it. We were testing things out. I've got changes to make and I'll send out a new PR soon.

Issue #23, opened 2021-11-13
Create a particle class framework


wkchuang:
Create a particle class, using pytype for static analysis and incorporating a library for physical units / conversions.

Issue #7, opened 2021-11-02
set up devcontainer and dockerfile


ngam:
- use ms template for now 
- rely on pip in container for now
- document procedure and document conda procedure as well

PR #27, opened 2021-11-23
Improved Particle class & methods, typing & units. 

wkchuang:
Added decorators to functions to conduct units checking.
Added docstrings to many (but not all) functions.
Added vsliveshare extension so we can hold pair programming/audio chat sessions.

Co-authored-by: aktwo <aktwo@users.noreply.github.com>

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,7 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.202.5/containers/python-3/.devcontainer/base.Dockerfile
 
 # [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
-ARG VARIANT="3"
+ARG VARIANT="3.9"
 FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
 
 # [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
--- a/aerosol_dynamics/particle.py
+++ b/aerosol_dynamics/particle.py
@@ -1,58 +1,129 @@
 import numpy as np
 from aerosol_dynamics import physical_parameters as pp
+from . import u
 
 class Particle:
-    def __init__(self, name: str, radius: float, density: float, charge: float):
-        self.__name = name
-        self.__radius = radius
-        # self.__density = density
-        self.__charge = charge
-        self.__mass = density * (4*np.pi/3) * (radius**3)
+    """Sets the class for creating particles.
+    This forms the framework for particle-particle and gas-particle interactions.
 
-    def name(self):
-        return self.__name
+    Attributes:
+        name (str): The name of the particle.
+        radius (float): The radius of the particle.
+        density (float): The density of the particle.
+        charge (float): The charge of the particle.
+        mass (float): The mass of the particle.
+    """
 
-    def mass(self):
-        return self.__mass
- 
-    def radius(self):
-        return self.__radius
+    def __init__(self, name: str, radius, density, charge):
+        '''
+        Constructs the particle object.
 
-    def charge(self):
-        return self.__charge
+        Parameters:
+            name (str): The name of the particle.
+            radius (float): The radius of the particle.
+            density (float): The density of the particle.
+            charge (float): The charge of the particle.
+        '''
+        self._name = name
+        self._radius = radius
+        # self._density = density
+        self._charge = charge
+        self._mass = density * (4*np.pi/3) * (radius**3)
 
+    def name(self) -> str:
+        '''Returns the name of the particle.'''
+        return self._name
+
+    @u.wraps(u.kg, [None])
+    def mass(self) -> float:
+        '''Returns mass of a particle. Checks units. [kg]'''
+        return self._mass
+
+    @u.wraps(u.m, [None])
+    def radius(self) -> float:
+        '''Returns radius of a particle. Checks units. [m]'''
+        return self._radius
+
+    @u.wraps(u.dimensionless, [None])
+    def charge(self) -> float:
+        '''Returns charge of a particle. Checks units. [unitless]'''
+        return self._charge
+
+    @u.wraps(u.dimensionless, [None])
     def knudsen_number(self) -> float:
+        '''Returns a particle's Knudsen number. Unitless.
+
+        The Knudsen number seeks to reflect the relative length scales of the
+        particle and the suspending fluid (air, water, etc.). This is calculated
+        by the mean free path of the medium divided by the particle radius.
+        '''
         return pp.MEAN_FREE_PATH_AIR / self.radius()
 
+    @u.wraps(u.dimensionless, [None])
     def slip_correction_factor(self) -> float:
+        '''Returns a particle's Cunningham slip correction factor. Unitless.
+
+        The slip correction factor is a dimensionless quantity that accounts for
+        non-continuum effects when calculating the drag on small particles.
+        This is a deviation from Stokes' Law; Stokes assumes a no-slip
+        condition that is no longer correct at high Knudsen numbers.
+        The slip correction factor is used to calculate the friction factor.
+        See Eq 9.34 in Atmos. Chem. & Phys. (2016) for more information.
+        '''
         knudsen_number: float = self.knudsen_number()
         return 1 + knudsen_number * (1.257 + 0.4*np.exp(-1.1/knudsen_number))
 
+    @u.wraps(u.kg / u.s, [None])
     def friction_factor(self) -> float:
-        slip_correction_factor: float = self.slip_correction_factor()
-        return (6 * np.pi * pp.MEDIUM_VISCOSITY * self.radius() / slip_correction_factor)
+        '''Returns a particle's friction factor. [N-s/m].
 
+        The friction factor is a property of the particle's size and the medium
+        that the particle is in. Multiplying the friction factor by the fluid
+        velocity gives the drag force on the particle.
+        '''
+        slip_correction_factor: float = self.slip_correction_factor()
+        return 6 * np.pi * pp.MEDIUM_VISCOSITY * self.radius() / slip_correction_factor
+
+    @u.wraps(u.kg, [None, None])
     def reduced_mass(self, other) -> float:
+        '''Returns the reduced mass of two particles. [kg].
+
+        The reduced mass is an "effective inertial" mass that allows a two body
+        problem to be solved as a one body problem.
+        '''
         return self.mass() * other.mass() / (self.mass() + other.mass())
 
     def reduced_friction_factor(self, other) -> float:
+        '''Returns the reduced friction factor between two particles. [N-s/m]
+       
+        Similar to the reduced mass, the reduced friction factor allows a two
+        body problem to be solved as a one body problem.
+        '''
         return (self.friction_factor() * other.friction_factor()
             / (self.friction_factor() + other.friction_factor()))
 
     def coulomb_potential_ratio(self, other) -> float:
+        '''
+        '''
         numerator = -1 * self.charge() * other.charge() * (pp.ELEMENTARY_CHARGE_VALUE ** 2)
         denominator = 4 * np.pi * pp.ELECTRIC_PERMITTIVITY * (self.radius() + other.radius())
         return numerator / (denominator * pp.BOLTZMANN_CONSTANT * pp.TEMPERATURE)
 
     def coulomb_enhancement_kinetic_limit(self, other) -> float:
+        '''
+        '''
         coulomb_potential_ratio = self.coulomb_potential_ratio(other)
         return 1 + coulomb_potential_ratio if coulomb_potential_ratio >=0 else np.exp(coulomb_potential_ratio)
 
     def coulomb_enhancement_continuum_limit(self, other) -> float:
+        '''
+        '''
         coulomb_potential_ratio = self.coulomb_potential_ratio(other)
         return coulomb_potential_ratio / (1 - np.exp(-1*coulomb_potential_ratio)) if coulomb_potential_ratio != 0 else 1
 
     def diffusive_knudsen_number(self, other) -> float:
+        '''
+        '''
         reduced_mass = self.reduced_mass(other)
         coulomb_enhancement_continuum_limit = self.coulomb_enhancement_continuum_limit(other)
         reduced_friction_factor = self.reduced_friction_factor(other)
@@ -61,6 +132,8 @@ class Particle:
             (reduced_friction_factor * (self.radius() + other.radius()) * coulomb_enhancement_kinetic_limit))
 
     def dimensionless_coagulation_kernel_hard_sphere(self, other) -> float:
+        '''
+        '''
         # Constants for the chargeless hard-sphere limit
         HSC1 = 25.836
         HSC2 = 11.211
@@ -69,17 +142,21 @@ class Particle:
 
         diffusive_knudsen_number = self.diffusive_knudsen_number(other)
 
-        numerator = ((4 * np.pi * diffusive_knudsen_number**2) + (HSC1 * diffusive_knudsen_number**3) + 
+        numerator = ((4 * np.pi * diffusive_knudsen_number**2) + (HSC1 * diffusive_knudsen_number**3) +
                     ((8 * np.pi)**(1/2) * HSC2 * diffusive_knudsen_number**4))
         denominator = (1 + HSC3 * diffusive_knudsen_number + HSC4 * diffusive_knudsen_number**2 + HSC2 * diffusive_knudsen_number**3)
 
         return numerator / denominator
 
     def collision_kernel_continuum_limit(self, other) -> float:
+        '''
+        '''
         diffusive_knudsen_number = self.diffusive_knudsen_number(other)
         return 4 * np.pi * (diffusive_knudsen_number**2)
 
     def collision_kernel_kinetic_limit(self, other) -> float:
+        '''
+        '''
         diffusive_knudsen_number = self.diffusive_knudsen_number(other)
         return np.sqrt(8 * np.pi) * diffusive_knudsen_number
 
--- a/aerosol_dynamics/particle.py
+++ b/aerosol_dynamics/particle.py
@@ -1,0 +1,107 @@
+import numpy as np
+from aerosol_dynamics import physical_parameters as pp
+
+class Particle:
+    def __init__(self, name: str, radius: float, density: float, charge: float):
+        self.__name = name
+        self.__radius = radius
+        # self.__density = density
+        self.__charge = charge
+        self.__mass = density * (4*np.pi/3) * (radius**3)
+
+    def name(self):
+        return self.__name
+
+    def mass(self):
+        return self.__mass
+ 
+    def radius(self):
+        return self.__radius
+
+    def charge(self):
+        return self.__charge
+
+    def knudsen_number(self) -> float:
+        return pp.MEAN_FREE_PATH_AIR / self.radius()
+
+    def slip_correction_factor(self) -> float:
+        knudsen_number: float = self.knudsen_number()
+        return 1 + knudsen_number * (1.257 + 0.4*np.exp(-1.1/knudsen_number))
+
+    def friction_factor(self) -> float:
+        slip_correction_factor: float = self.slip_correction_factor()
+        return (6 * np.pi * pp.MEDIUM_VISCOSITY * self.radius() / slip_correction_factor)
+
+    def reduced_mass(self, other) -> float:
+        return self.mass() * other.mass() / (self.mass() + other.mass())
+
+    def reduced_friction_factor(self, other) -> float:
+        return (self.friction_factor() * other.friction_factor()
+            / (self.friction_factor() + other.friction_factor()))
+
+    def coulomb_potential_ratio(self, other) -> float:
+        numerator = -1 * self.charge() * other.charge() * (pp.ELEMENTARY_CHARGE_VALUE ** 2)
+        denominator = 4 * np.pi * pp.ELECTRIC_PERMITTIVITY * (self.radius() + other.radius())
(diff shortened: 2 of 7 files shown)


aerosol_dynamics/particle_test.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+from aerosol_dynamics import particle
+from . import ureg
```
ngam, line 3: I have heard read we should we avoid any relative imports. So we may need to change this in the future. We also should leave `__init__.py` as empty as possible.

aerosol_dynamics/physical_parameters.py
```diff
@@ -0,0 +1,29 @@
...
+TEMPERATURE = 300 * ureg.K
+
+# elementary charge in C
+ELEMENTARY_CHARGE_VALUE = 1.60217662e-19 * ureg.C
+
+# dielectric constant of air
+EPSI = 1.0005 # unitless
+
+# permittivity of free space in F/m
+EPS0 = 8.85418782e-12 * ureg.F / ureg.m
```
ngam, line 26: Interesting, does this unit thing follow the legit conventions of uppercase, lowercase, etc. when it comes to units?

As an aside, I will also update some of the constants above to follow the latest iteration. Apparently, in 2019, the standard unit committee (or whatever they're called) changed the definitions of basic units to rely more on fundamental constants. So currently, there is an exact number for `BOLTZMANN_CONSTANT` and `ELEMENTARY_CHARGE_VALUE` and I think they're actually almost exactly the same as above (please change the power on k)

`BOLTZMANN_CONSTANT = 1.380649e-23`
`ELEMENTARY_CHARGE_VALUE = 1.602176634e-19`

If interested, see more about the politics of units: https://en.wikipedia.org/wiki/2019_redefinition_of_the_SI_base_units 

We should probably eventually redefine everything to go back to the very basic constants; that will take some time, but it will be an interesting exercise.

aerosol_dynamics/particle_test.py
```diff
@@ -0,0 +1,50 @@
...
+particle2 = particle.Particle(name='particle2',
+                              radius=1.0e-7 * ureg.m,
+                              density=1.8 * ureg.kg / ureg.m**3,
+                              charge=1.0 * ureg.C,)
+
+particle3 = particle.Particle(name='particle3',
+                              radius=1.0e-7 * ureg.lb,
+                              density=1 * ureg.kg / ureg.m**3,
+                              charge=3,)
+
```
ngam, line 19: Generally, I think this should be rewritten like this:

```python
particle2 = particle.Particle(
    name='particle2', # one can add comments like this
    radius=1.0e-7 * ureg.m,
    density=1.8 * ureg.kg / ureg.m**3,
    charge=1.0 * ureg.C, # I think we shouldn't have space before/after * and /
)
```

Also probably use double quotations, `"`, here and everywhere to be consistent; Python treats them the same.

aerosol_dynamics/particle.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
```
ngam, line 1: The level of documentation is insufficient; however, given that I wrote the previous iteration of these calculations, I will be happy to step in and add documentation after the fact. Next time, we will be blocking merge requests for any documentation violations --- hopefully automatically by the linter ;)

aerosol_dynamics/particle.py
```diff
@@ -0,0 +1,107 @@
...
+from aerosol_dynamics import physical_parameters as pp
+
+class Particle:
```
ngam, line 4: I will wait to judge this specific `class` treatment of particles until later... It may be end up too limiting

aerosol_dynamics/particle_test.py
```diff
@@ -0,0 +1,50 @@
...
+def reduced_mass_test():
+    reduced_mass_1_2 = particle1.reduced_mass(particle2)
+    assert reduced_mass_1_2 == pytest.approx(4.18879e-27)
+    assert reduced_mass_1_2.check('[mass]')
+reduced_mass_test()
+
+def reduced_friction_factor_test():
+    reduced_friction_factor_1_2 = particle1.reduced_friction_factor(particle2)
+    assert reduced_friction_factor_1_2 == pytest.approx(1.833335)
+    assert reduced_friction_factor_1_2.check('[None]')
```
ngam, line 50: One thing about these tests going forward: In my mind a really important test is going to be related to shape/size. Say you give a coagulation routine two vectors of particles, what does it return? We need to have predefined rules for that (if two vectors, it returns a matrix of their lengths). Generally, we could perhaps abstract all quantities in the entire code to some tensor/matrix definitions so that we can make use of Python broadcasting features more readily. I don't think that's very important yet, but I wanted to note it.

ngam (changes requested):
Thanks a lot for this!

- [ ] Double check the physical parameters to ensure they are accurate.
- [ ] Documentation is lacking, but I volunteer to take care of it after the fact.
- [ ] Consider changing the style in writing functions.

aerosol_dynamics/particle_test.py
wkchuang: I'll see what I can do about these imports.
__init__.py was necessary due to how pint works. Each file that imports pint creates a separate units registry, and the registries do not talk to each other. If we pass a radius from particle.py to particle_test.py and try to match it with a radius created in particle_test.py, pint will say the two radii have different units registries. As our file structure evolves with the project, that may be an issue we have to resolve.

aerosol_dynamics/particle.py
wkchuang: That's odd, I definitely added documentation here; this appears to be the old pull request! Did I miss something? 
I tried to find information on the charge_ratio, digging into your research letter, but couldn't make enough sense of it to add documentation.

ngam: No problem if you need to redo, and don't worry about documentation --- don't let that slow you down. I am a pretty verbose person and happy to do more of the documentation :)

ngam: My proposal: Even if this is the old pr, don't worry about it. You can do another pr later. Let's work to get this through first.

aerosol_dynamics/particle_test.py
wkchuang: Single quotations are used consistently throughout my files because that's my style; it feels cleaner to me. I'll change it if others are used to double quotations. 

I think I prefer keeping the spaces between * and /, only to space it out more. It can become harder to read when units get complicated. We may need to cut out spaces in those cases and group units together

aerosol_dynamics/particle_test.py
wkchuang: That's a good note, and I am keeping that in mind too. We'll figure out a solution once we have the framework set up and start adding dynamics.

aerosol_dynamics/particle_test.py
ngam: We can go either way with quotes --- we should try to stick to one. I like the idea of double quotes because it is clearer, and it encourages me to use less of them, while a single quote can blend in weirdly with many other things.

For spacing, I would follow [this](https://www.python.org/dev/peps/pep-0008/#other-recommendations), which I find especially helpful when debugging and disentangling the order of operations

aerosol_dynamics/physical_parameters.py
wkchuang: Can you elaborate about changing the power on k? 
From my understanding, the values themselves have not changed, only the way they're defined. 
Pint should sort out most units issues for us. We'll have to do some testing to make sure certain combinations of units remains the way we want, but generally speaking Pint will take care of any conversions, and we can leave our constants in the form they are most commonly used/understood (e.g. no need to rewrite Coulomb as Ampere * seconds)

ngam: It should be `BOLTZMANN_CONSTANT = 1.380649e-23...` not `BOLTZMANN_CONSTANT = 1.380649e-21 * ureg.m ** 2 * ureg.kg / ureg.s / ureg.K` --- that is, Boltzmann constant (k) has a power of -23, not -21. 

Sorry, I should have said the power of Boltzmann constant, not "k", when we have many ks around...

ngam: (isn't that what the thing atop the 10 is called? the number 10^5 has a power of 5... or am I making this up?)

aerosol_dynamics/particle_test.py
wkchuang: That's fine, we can go with double quotes. I'll make those changes. 

The reference you pointed to is what I'll use when there are mixed order of operations, like * / and + -. Both PEP8 and the Google style guide are unclear about what to do in a sequence of * and /. It does feel a little strange having spaces given the lack of space around the =, so I'm amenable to removing them.

aerosol_dynamics/physical_parameters.py
wkchuang: "Power" is fine, "Exponent" might be clearer. I thought you meant ureg.K had the wrong exponent and was confused. Good catch on the number though

aerosol_dynamics/particle_test.py
ngam: No need to follow right away, just thoughts for now. We will see what other people think/prefer. It won't make much of a difference eventually. What I have been doing is just following whatever the defaults on VSC are saying, unless something that deeply bothers me 😄

.devcontainer/Dockerfile
```diff
@@ -1,21 +1,21 @@
...
-
-# [Optional] Uncomment this section to install additional OS packages.
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get -y install --no-install-recommends vim
-
-# [Optional] Uncomment this line to install global node packages.
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.202.5/containers/python-3/.devcontainer/base.Dockerfile
+
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+ARG VARIANT="3.9"
```
ngam, line 4: Good call using 3.9 for the variant (at least for me, it would have otherwise defaulted to 3.10(-bullseye), which is still not well supported). However, I think we can don't need to actually repeat the variant here; it is already defined in the `devcontainer.json` file actually (line 13). No problem keeping it though for now :)

aerosol_dynamics/physical_parameters.py
```diff
@@ -2,28 +2,32 @@
...
-MEDIUM_VISCOSITY = 18.27e-6 * ureg.kg / ureg.m / ureg.s
+MEDIUM_VISCOSITY = 18.27e-6 * u.kg / u.m / u.s
 
 # mean free path of air in m
-MEAN_FREE_PATH_AIR = 65e-9 * ureg.m
+MEAN_FREE_PATH_AIR = 65e-9 * u.m
 
 # Boltzmann's constant in m^2 kg s^-2 K^-1
-BOLTZMANN_CONSTANT = 1.380649e-21 * ureg.m ** 2 * ureg.kg / ureg.s / ureg.K
+BOLTZMANN_CONSTANT = 1.380649e-21 * u.m ** 2 * u.kg / u.s / u.K
```
ngam, line 14: This need to be `e-23`. Other than that, I approve of this pr, stellar work 🥇 

Thank you both!!!

ngam (changes requested):

.devcontainer/Dockerfile
ngam: Actually never mind --- we should likely want to keep both. Sorry about the confusion.

aerosol_dynamics/physical_parameters.py
wkchuang: Thanks for catching that (again!)

ngam:
@wkchuang + @Gorkowski do you prefer another review or should I let this go through?

wkchuang:
I'd like to give @Gorkowski a chance to review it, so we're all on the same page on our first step.

aerosol_dynamics/particle.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        slip_correction_factor: float = self.slip_correction_factor()
+        return 6 * np.pi * pp.MEDIUM_VISCOSITY * self.radius() / slip_correction_factor
+
+    @u.wraps(u.kg, [None, None])
+    def reduced_mass(self, other) -> float:
+        """Returns the reduced mass of two particles. [kg].
+
+        The reduced mass is an "effective inertial" mass that allows a two body
+        problem to be solved as a one body problem.
```
Gorkowski, line 92: what does that two body to one body mean? (in our context) I think we'll have to explain this in an example eventually.

aerosol_dynamics/particle.py
```diff
@@ -0,0 +1,184 @@
...
+       
+        Similar to the reduced mass, the reduced friction factor allows a two
+        body problem to be solved as a one body problem.
+        """
+        return (self.friction_factor() * other.friction_factor()
+            / (self.friction_factor() + other.friction_factor()))
+
+    def coulomb_potential_ratio(self, other) -> float:
+        """
+        """
```
Gorkowski, line 107: are these in S&P, or do we need another citation for these charge equations.

aerosol_dynamics/physical_parameters.py
```diff
@@ -0,0 +1,33 @@
+"""
+Centralized location for physical parameters.
```
Gorkowski, line 2: This is good for now, but I think some things (temp, viscosity, mean free path) will be removed from this when we make an 'air parcel' class for the particles and gases to be in.

Gorkowski:
@wkchuang Nice work, this whole unit thing seems cool!!

ngam:
Closes #7 
Closes #23 

(so that these issues are automatically closed when this is merged)

ngam (approved):

aerosol_dynamics/particle.py
ngam: We don't strictly speaking need another citation for these things. I can later do a thorough documentation of all equations step-by-step. These things are sort of well known --- a simplistic treatment along these lines does exists in S&P (I believe in he appendices). However, this particular treatment takes a detour into the dimensionless direction --- quite different than classic methods, but still well-documented and studied in the literature. In the future, I can add other alternative approaches as backup or options.

aerosol_dynamics/particle.py
ngam: If you look here, you will see we are not using the two masses, instead just the reduced mass. Maybe a more practical way to look at this would be to consider the reduced mass as a correction of the size effects on the second line.  

https://github.com/uncscode/particula/blob/3c978f7ae83d166cb12fd6f63526e37744d115fb/aerosol_dynamics/particle.py#L131-L132

ngam: Btw, I would consider rewriting this example like this in the future (no need to rewrite it now --- we will likely need to rewrite it later anyway for corrections; I think I have made a mistake or two somewhere in the calculations that the tests will later help us figure out; this calculation doesn't exactly coincide with the values I have in mind)

```python
    return (
        (
            pp.TEMPERATURE*pp.BOLTZMANN_CONSTANT*reduced_mass**0.5
        )*coulomb_enhancement_continuum_limit/(
            reduced_friction_factor*(self.radius() + other.radius())
        )*coulomb_enhancement_kinetic_limit
    ) 
```

ngam: Okay, just a little more context on the "reduced mass" thingy:

Usually you could get away with assuming the smaller particle is mobile while the bigger one is stationary, thus a coagulation problem is really just a slightly modified condensation problem or vice versa. And that actually works remarkably well, except when the particles have similar sizes (or volumes or masses) and especially in the free molecular regime (i.e. both particles are small enough compared to the surrounding gas). Essentially, that's really what's hard about coagulation --- you've got to link the two relatively well-understood limits through the transition regime (both in terms of particle--gas interactions as well as particle--particle interactions). So you start taking into account all higher-order dependencies because they actually matter. The reduced mass definition goes like (m1 x m2)/(m1 + m2), let's say m1 > m2, and remember that people often treat the smaller particle as mobile and the bigger particle as stationary, then you have usually just take the mass in question as m2 (the smaller mass). The reduced mass is just a correction on that. If m1 >> m2, then (m1 x m2)/(m1 + m2) = 1 (approx); that is, the correction is negligible. But if the masses are similar, then, this correction can be significant (the correction is always equal to or less than 1: m1/(m1+m2) <= 1; hence "reduced" in reduced mass, the mass of the smaller particle is effectively reduced). 

What does this actually mean? It just means both are moving and both are contributing to the collision --- remember...

wkchuang: Just me venting about how complex equations don't translate efficiently into coding style: 
Making long equations readable is something I'm still trying to get comfortable with. I tend to read equations horizontally, so parsing equations vertically make it harder for me to get the big picture. Similarly, splitting open and close parentheses on different lines breaks the visual "this is a group" that parentheses are supposed to give. In addition, in the example I would want the division sign / should be more prominent, because it effectively splits the expression in two parts. In perhaps conventional coding style, that is not as clear. I understand the desire to limit # characters in a line, but I feel it can be restrictive at times and not allow for capturing complex equations well. 

If we have to rewrite it to fit our style guide though, then we'll rewrite it.

wkchuang: Adding an example is a good idea @Gorkowski! I only added a simple explanation that I pulled from wiki, but it's worth painting a clearer picture in our documentation. My hope is that a first year grad student would be able to go through this and understand almost everything on a surface level without having to refer to outside texts or googling (and perhaps some forgetful phd students too), though I'm open to having more or less explanations.

aerosol_dynamics/particle.py
wkchuang: I think it would be good to have references for where each equation comes from. Even if it's well-documented in the literature, I would hate for a new student (or even me, who never worked with charges) to have to dig for papers to convince themselves it's correct or understand the context for how it's derived and used. I couldn't find this exact equation in S&P or through a cursory search for "coulomb potential ratio", so having it explicitly pointed out would be tremendously helpful.

aerosol_dynamics/physical_parameters.py
wkchuang: Aktwo and I started working on an Environment class that would attempt to address this. Temp/Pressure would move there, and I guess we would put mean free path there to be calculated from environment properties. I can address that in a separate issue.

aerosol_dynamics/particle.py
ngam: > splitting open and close parentheses on different lines breaks the visual "this is a group" that parentheses are supposed to give

quite the opposite actually, no? Which one of these asserts the groupness of expressions between the parentheses more prominently?

```python
(
    five*time*twnetytwo + fifty*times*twothousand
)/(
    something_else
)
```

```python
(five*time*twnetytwo + fifty*times*twothousand)/
(something_else)
```

```python
(five*time*twnetytwo + fifty*times*twothousand)/(something_else)
```

I don't think what you're doing is wrong --- it is just my opinion that writing equations vertically makes it easier to debug them. And this happens to get me around the line limits and all that stuff imposed by the linter.

aerosol_dynamics/particle.py
ngam: On "coulomb potential ratio" there is a high chance I just made that up to make the whole logic flow better. I will submit an issue now to add a more thorough documentation of these steps.

aerosol_dynamics/physical_parameters.py
ngam: @aktwo @wkchuang it would be good to make that public whenever you feel comfortable --- either in your forks or in a remote branch here. Up to you though :)

ngam:
#28 to improve docs, etc.

aerosol_dynamics/particle.py
wkchuang: I think it's just what my brain is used to, which does more visual grouping. The separating of parentheses and extra lines is like extra whitespace to me -- distracting (unless it serves to separate nested parentheses, which are visually confusing!). There's sort of a grey area where we transition from horizontal to vertical, right? -- we probably wouldn't split something short like:
(
    argone + argtwo
) * argthree
because it's easier to read as: (argone+argtwo) * argthree. (To me at least!)

And I would prefer:
(first group) * factor_one / 
(
    (
        (second group) + 
        (third group)
    ) * factor_two
)
over:
(
    first group
) * factor_one / 
(
    (
        (
            second group
        ) + 
        (
            third group
        )
    ) * factor_two
)

BUT, I am just elaborating on my preferences, which is amenable to change and adaptation! Our style guide is silent on it, and it appears to have no clear consensus online, so I think it comes down to what we are comfortable with and what communicates well. I don't want to be strictly bound to an 80 character line limit because a linter says so (and wasn't developed with math equations in mind) when we have descriptive variable names. Sure it's good to try to keep it shorter most of the time, but splitting too much vertically can be more difficult to understand too. I think I sound more stubborn than I feel, so I want to make it clear that I'm open to a new style. Who knows, as we write more equations, I might find that my style is not great for some of them. Any thoughts or preferences, @Gorkowski?

ngam: I would say the following on the 80-col (or 72-col) limit: If you can follow it, it is really good and make the code more portable (i.e. easier to view in multiple editors like vim, vscode, github.dev, etc.)

But I wouldn't go out of my way to break a long link jus to satisfy. However, I would break a long path (../.../..../..../..../...) to satisfy it. I have recently started reading some old-school programming stuff and I am influenced by them (the unix philosophy, anyone? _awk & sed_ has been fun!)

The idea here is not to be stubborn and picky, and I would really hate for this stuff to delay us. If you give me your code with no formatting and it works, I am happy with it and I can volunteer to lint it --- or just leave it as is. At the end of the day, I don't want any of these random extras to derail our progress!

ngam: And I think what matters most is 1) consistency within one document and 2) consistency in code overall --- in this order.

wkchuang: Fair enough! Well, I think I'm happy to switch it to a more vertical profile. I question how much old-school stuff still applies (e.g. monitors are wider these days, though I realize laptop screens are still small). But you are absolutely right, I wouldn't want this to prevent progress on the more substantial areas either. Refactoring is *always* an option later. I'll change some of these expressions and push another version tonight.

ngam: > I'll change some of these expressions and push another version tonight.

Please don't. Let's worry about that later. This is more than good enough for now. We can revisit. And don't worry, I am outspoken and I bring things up, so you will hear more from me :)

PR #63, opened 2021-12-22
improving github action workflows 

### LINKED ISSUE #54: improve workflows
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

1. split linters (flake8 and pylint)
2. make workflow fail on flake8 linter
3. provide better titles, and documentation
4. consider eliminating manyversions from linters & maybe build?

**Describe alternatives you've considered**

**Additional context**

ngam:
Fixes #54

Highlights:
- added codeowners (for now everywhere, more granual later)
- split linters
- better titles of workflows
- make flake8 fail if errors
- removed manyversions from linters

Notes (if applicable):
- n/a

### DIFF
--- a/.github/workflows/jupyter-book.yml
+++ b/.github/workflows/jupyter-book.yml
@@ -1,5 +1,5 @@
 # from: https://jupyterbook.org/publish/gh-pages.html
-name: jupyter-book
+name: jupyter-book deployment
 
 # Only run this when the master branch changes
 on:
--- a/.github/workflows/jupyter-book.yml
+++ b/.github/workflows/jupyter-book.yml
@@ -1,5 +1,5 @@
 # from: https://jupyterbook.org/publish/gh-pages.html
-name: jupyter-book deployment
+name: jupyter-book
 
 # Only run this when the master branch changes
 on:
@@ -15,7 +15,7 @@ on:
 
 # This job installs dependencies, builds the book, and pushes it to `gh-pages`
 jobs:
-  deploy-book:
+  deployment:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package-flake8.yml
+++ b/.github/workflows/python-package-flake8.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: flake8
+name: flake8 linting
 
 on:
   push:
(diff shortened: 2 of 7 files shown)


.github/workflows/python-package-flake8.yml
```diff
@@ -0,0 +1,37 @@
...
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest pylint
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        # ngam, 2021-12-21: removed `--exit-zero` to make it fail
```
ngam, line 36: @wkchuang essentially, this edit (taking out `--exit-zero` makes flake8 fail for the slightest error. Previously, it was lenient in that it would just report them as warnings, but not fail the build unless there is a dramatic error (line 34). We can try this for a while and see how if we want to customize it (there are two ways: 1) through a flake8 dotfile or 2) in this call right here).

.github/workflows/jupyter-book.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 # This job installs dependencies, builds the book, and pushes it to `gh-pages`
 jobs:
-  deploy-book:
+  deployment:
```
wkchuang, line 18: Just for my understanding, what does this change do?

ngam: Nothing, just the titles of the workflows, e.g. pytest / testing (3.7) (pull_request) Successful in 38s or see this screenshot: 
Screen Shot 2021-12-21 at 21 32 55

ngam: Note this particular one (jupyter-book) is only triggered if the /docs/ folder changes (after merging). These tests actually run twice: once here in the PR and once right after the PR, so the jupyter-book one is usually done only after

.github/workflows/python-package-flake8.yml
wkchuang: Cool, that's good to know, thanks!

.github/workflows/jupyter-book.yml
ngam: Needless to say, I am happy to change the wording. I just wanted to make them as short and descriptive as possible

.github/workflows/python-package-pytest.yml
```diff
@@ -10,13 +10,13 @@ on:
     branches: [ main ]
 
 jobs:
-  build:
+  testing:
```
wkchuang, line 13: What does changing this from `build` to `testing` do? (I checked the link on top but I think this change goes deeper than their README. A quick explanation or link please?)

ngam: Same as above, in this particular example, the title of the workflow (as it would appear under actions tab) would be "pytest / testing (3.7) (pull_request)" where "pytest" comes from line 4, "testing" from this line (13), 3.7 is part of the version matrix in line 18, and pull_request from line 9.

Before editing the title would be: "Python package test / build (3.7) (pull_request)" (for example, see here: https://github.com/uncscode/particula/runs/4602117804?check_suite_focus=true


More details in general here: https://github.com/uncscode/particula/actions

wkchuang (approved):

PR #84, opened 2022-01-01
Wkchuang/issue82

### LINKED ISSUE #82: Inputs without units to Environment and Particle Classes are accepted
Environment and Particle classes are accepting inputs without designating units.
Inputs are somehow then assigned units to pass the class tests.

wkchuang:
Fixes #82

Highlights:
- Classes will now check inputs for the correct units.
- Temperature inputs will automatically be converted to base units (Kelvin) prior to any mathematical operations. 

Notes (if applicable):
- A UnitStrippedWarning pops up when running `particle_test`. Need help diagnosing the issue.
-

### DIFF
--- a/particula/__init__.py
+++ b/particula/__init__.py
@@ -1,4 +1,4 @@
-""" a simple, fast, and powerful particle simulator.
+""" A simple, fast, and powerful particle simulator.
 
     particula is a simple, fast, and powerful particle simulator,
     or at least two of the three, we hope. It is a simple particle
@@ -18,5 +18,12 @@ from pint import UnitRegistry
 
 # u is the unit registry name.
 u = UnitRegistry()
+
+# Temperature has the quirky issue that mathematical operations on temperature
+# are problematic. e.g. 298.15K + 273.15K != 25C + 0C.
+# The following setting converts all offset units to base units prior to
+# any operations.
+# https://pint.readthedocs.io/en/latest/nonmult.html
+u.autoconvert_offset_to_baseunit = True
 
 __version__ = '0.0.4'
--- a/particula/aerosol_dynamics/environment.py
+++ b/particula/aerosol_dynamics/environment.py
@@ -1,4 +1,4 @@
-"""environment class
+"""Environment class
 """
 
 import numpy as np
@@ -12,12 +12,13 @@ class Environment:
     with properties such as temperature and pressure
     and derived properties such as air viscosity.
     """
+    @u.wraps(None, [None, u.degK, u.Pa])
     def __init__(self, temperature, pressure):
         """Function calls for enviornment class."""
         self._temperature = temperature
         self._pressure = pressure
 
-    @u.wraps(u.K, [None])
+    @u.wraps(u.degK, [None])
     def temperature(self) -> float:
         """Returns the temperature of the environment."""
         return self._temperature
@@ -35,8 +36,8 @@ class Environment:
         Sutherland Viscosity Law.
         """
         mu_ref = 1.716e-5 * u.Pa * u.s  # Viscosity at T_REF
-        t_ref = 273.15 * u.K
-        suth_const = 110.4 * u.K  # Sutherland constant
+        t_ref = 273.15 * u.degK
+        suth_const = 110.4 * u.degK  # Sutherland constant
 
         return (
             mu_ref *
--- a/particula/aerosol_dynamics/particle.py
+++ b/particula/aerosol_dynamics/particle.py
@@ -1,4 +1,4 @@
-""" to instantiate particles and calculate their properties.
+""" To instantiate particles and calculate their properties.
 
     This module contains the Particle class, which is used to
     instantiate the particles and calculate their properties.
@@ -47,6 +47,7 @@ class Particle:
         mass    (float) [kg]
     """
 
+    @u.wraps(None, [None, None, u.m, u.kg / u.m**3, None])
     def __init__(self, name: str, radius, density, charge):
         """Constructs particle objects.
 
--- a/particula/aerosol_dynamics/tests/environment_test.py
+++ b/particula/aerosol_dynamics/tests/environment_test.py
@@ -1,4 +1,4 @@
-"""tests for the environment module.
+"""Tests for the environment module.
 """
 
 import pytest
@@ -6,8 +6,8 @@ from particula import u
 from particula.aerosol_dynamics import environment
 
 standard_environment = environment.Environment(
-    temperature=298,  # * u.K,
-    pressure=101325,  # * u.Pa,
+    temperature=298 * u.degK,
+    pressure=101325 * u.Pa,
 )
 
 
@@ -15,7 +15,7 @@ def test_getters():
     """
     Tests the getters for the Environment class.
     """
-    assert standard_environment.temperature() == 298 * u.K
+    assert standard_environment.temperature() == 298 * u.degK
     assert standard_environment.pressure() == 101325 * u.Pa
 
 
--- a/particula/aerosol_dynamics/tests/parcel_test.py
+++ b/particula/aerosol_dynamics/tests/parcel_test.py
@@ -1,4 +1,4 @@
-"""tests for the parcel module.
+"""Tests for the parcel module.
 """
 
 from particula import u
@@ -6,15 +6,15 @@ from particula.aerosol_dynamics import environment, parcel, particle
 
 small_particle = particle.Particle(
     name="small_particle",
-    radius=1.0e-9,
-    density=1.0,
+    radius=1.0e-9 * u.m,
+    density=1.0 * u.kg / u.m**3,
     charge=1,
 )
 
 large_particle = particle.Particle(
     name="large_particle",
-    radius=1.0e-7,
-    density=1.8,
+    radius=1.0e-7 * u.m,
+    density=1.8 * u.kg / u.m**3,
     charge=1,
 )
 
--- a/particula/aerosol_dynamics/tests/particle_test.py
+++ b/particula/aerosol_dynamics/tests/particle_test.py
@@ -146,7 +146,7 @@ def test_dimensioned_coagulation_kernel():
 
     assert small_particle.dimensioned_coagulation_kernel(
         large_particle, standard_environment
-    ) == pytest.approx(2.738e-10 * u.m**3 / u.s, rel=10)
+    ) == pytest.approx(2.738e-10, rel=10)
     assert small_particle.dimensioned_coagulation_kernel(
         large_particle, standard_environment
     ).check("[length]**3/[time]")
@@ -163,7 +163,7 @@ def test_dimensioned_coagulation_kernel():
 
     assert negative_ion.dimensioned_coagulation_kernel(
         positive_particle, standard_environment_ip
-    ) == pytest.approx(1e-12 * u.m**3 / u.s, rel=10)
+    ) == pytest.approx(1e-12, rel=10)
     # rel=10 means an order of magnitude
     assert negative_ion.dimensioned_coagulation_kernel(
         positive_particle, standard_environment_ip
--- a/particula/aerosol_dynamics/tests/particle_test.py
+++ b/particula/aerosol_dynamics/tests/particle_test.py
@@ -2,7 +2,6 @@
 """
 
 import numpy as np
-import pint
 import pytest
 from particula import u
 from particula.aerosol_dynamics import environment, particle
@@ -19,13 +18,6 @@ large_particle = particle.Particle(
     radius=1.0e-7 * u.m,
     density=1.8 * u.kg / u.m**3,  # need fixing, add e3
     charge=1,
-)
-
-invalid_particle = particle.Particle(
-    name="invalid_particle",
-    radius=1.0e-7 * u.lb,
-    density=1 * u.kg / u.m**3,
-    charge=3 * u.C,
 )
 
 standard_environment = environment.Environment(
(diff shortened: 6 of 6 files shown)


ngam:
@wkchuang I think a more desirable outcome is:

- if inputs units are NONE (i.e. if user doesn't provide units, e.g. `std_env = environment.Environment(300, 1e5)`), then auto convert to base units
- if inputs units are defined (i.e. if user does provide units, e.g. `std_env = environment.Environment(300 * u.K, 1e5 * u.Pa)`), then take these units as is

Is it possible to implement such logic package-wide?

ngam:
> Classes will now check inputs for the correct units.

Should confirm this with a test if you still want to adopt this strategy. Try defining something without units and see if it fails.

wkchuang:
> > Classes will now check inputs for the correct units.
> 
> Should confirm this with a test if you still want to adopt this strategy. Try defining something without units and see if it fails.

I'll see if I can define a test. It failed when I tried running an input without units, I just have to figure out how to catch that and prevent the whole code from stopping.

ngam:
> I'll see if I can define a test. It failed when I tried running an input without units, I just have to figure out how to catch that and prevent the whole code from stopping.

There should be `assert == fail` type of thing in pytest. Let me check.

wkchuang:
> @wkchuang I think a more desirable outcome is:
> 
>     * if inputs units are NONE (i.e. if user doesn't provide units, e.g. `std_env = environment.Environment(300, 1e5)`), then auto convert to base units
> 
>     * if inputs units are defined (i.e. if user does provide units, e.g. `std_env = environment.Environment(300 * u.K, 1e5 * u.Pa)`), then take these units as is
> 
> 
> Is it possible to implement such logic package-wide?
> 
> Edit: how can we ensure that the user-provided `u` is the same as the one used by the package in this call `std_env = environment.Environment(300 * u.K, 1e5 * u.Pa)`? I don't think this is actually possible. Every invocation of `UnitRegistry` ought to be separate and I suppose we cannot easily ensure they are the same. Even worse, the user will likely find it quite hard to maintain one and only one `u`. Therefore, I think we should never expose `u` to the user --- it is just a backend for the package to check units, etc.

I think we should force the user to give valid inputs. Otherwise they only realize their inputs are wrong when they get the results. We would also have to assume there's a "default wrong unit" that we convert from (e.g. Is 32 in degF or degC?). The previous code added units from the subsequent units checks (Kelvin and Pa) --that's why the tests were passing even though inputs did not have units. But we wanted to make it explicit so people aren't entering 32 and find out they got results for 32 Kelvin.

I'm not sure what you mean by "every invocation of `UnitRegistry` ought to be separate". If they're not in a single instance of a UnitRegistry, then they clash and we'd also have to deal with lots of UnitRegistries. The point of one `u` is that mathematical operations work between them: you can multiply meters and seconds and then later divide by seconds. 
I'm also not sure how it is "hard to maintain one `u`", can you explain further?

ngam:
> I'm not sure what you mean by "every invocation of UnitRegistry ought to be separate". If they're not in a single instance of a UnitRegistry, then they clash and we'd also have to deal with lots of UnitRegistries. The point of one u is that mathematical operations work between them: you can multiply meters and seconds and then later divide by seconds.
I'm also not sure how it is "hard to maintain one u", can you explain further?

How will the user give us units? They will need access to `u`, right? They can either set their own `u` (i.e. `import particula; import pint; from pint import UnitRegistry; u = UnitRegistry()`) or they will have to use our `u` (i.e. `from particula import u` or `import particula; u = particula.u`). Can we make sure we are expecting the same `u` from the user as our package is defining?

Additionally, what if the user happens to import `pint` for a separate package/project they're using in the same environment?

ngam:
The only way to satisfy all these concerns is to house our own implementation of `pint` via submodules like I described many weeks ago. I could take care of that, but let's first think harder about this and not dig a hole for ourselves.

wkchuang:
> > I'm not sure what you mean by "every invocation of UnitRegistry ought to be separate". If they're not in a single instance of a UnitRegistry, then they clash and we'd also have to deal with lots of UnitRegistries. The point of one u is that mathematical operations work between them: you can multiply meters and seconds and then later divide by seconds.
> > I'm also not sure how it is "hard to maintain one u", can you explain further?
> 
> How will the user give us units? They will need access to `u`, right? They can either set their own `u` (i.e. `import particula; import pint; from pint import UnitRegistry; u = UnitRegistry()`) or they will have to use our `u` (i.e. `from particula import u` or `import particula; u = particula.u`). Can we make sure we are expecting the same `u` from the user as our package is defining?
> 
> Additionally, what if the user happens to import `pint` for a separate package/project they're using in the same environment?

If I'm understanding the premise correctly, then when users create their own `u` or import pint separately, those registries would clash with our `u` and throw an error if they ever meet. If they're building on our work, they should quickly clash. We can leave clear instructions on following our units process and a warning on the error.

ngam:
> > How will the user give us units? They will need access to `u`, right? They can either set their own `u` (i.e. `import particula; import pint; from pint import UnitRegistry; u = UnitRegistry()`) or they will have to use our `u` (i.e. `from particula import u` or `import particula; u = particula.u`). Can we make sure we are expecting the same `u` from the user as our package is defining?
> > Additionally, what if the user happens to import `pint` for a separate package/project they're using in the same environment?
> 
> If I'm understanding the premise correctly, then when users create their own `u` or import pint separately, those registries would clash with our `u` and throw an error if they ever meet. If they're building on our work, they should quickly clash. We can leave clear instructions on following our units process and a warning on the error.

Yes, but that would introduce a lot of potential restrictions (i.e. what if the user wants to use `pint` separately or with another package?). An easier solution is to just assume everything comes in standard units from the user. I personally view units as more of a testing mechanism rather than a full-fledged user-facing mechanism. Who's in our audience here? People interested in using the underlying calculations or people who are interested in a super-tight unit system?

particula/aerosol_dynamics/environment.py
```diff
@@ -12,12 +12,13 @@ class Environment:
     with properties such as temperature and pressure
     and derived properties such as air viscosity.
     """
+    @u.wraps(None, [None, u.degK, u.Pa])
```
ngam, line 15: Is this the key addition here?

particula/__init__.py
```diff
@@ -19,4 +19,11 @@
 # u is the unit registry name.
 u = UnitRegistry()
 
+# Temperature has the quirky issue that mathematical operations on temperature
+# are problematic. e.g. 298.15K + 273.15K != 25C + 0C.
+# The following setting converts all offset units to base units prior to
+# any operations.
+# https://pint.readthedocs.io/en/latest/nonmult.html
+u.autoconvert_offset_to_baseunit = True
```
ngam, line 27: Can you explain how this relates to `u.K` vis-a-vis `u.degK`?

ngam: Or is `u.K` simply incorrect?

particula/aerosol_dynamics/environment.py
ngam: If so, let's consider an alternative implementation:

```diff
-    @u.wraps(None, [None, u.degK, u.Pa])
     def __init__(self, temperature, pressure):
-        """Function calls for enviornment class."""
+        """ Function calls for environment class.
+          
+            Checks if the user provides a unit; if so, take to base units.
+            If user doesn't provide a unit, assume base units.
+        """
-        self._temperature = temperature
-        self._pressure = pressure
+        if type(temperature) == u.Quantity:
+            self._temperature = temperature.to_base_units()
+        else:
+             self._temperature = temperature * u.degK
+        if type(pressure) == u.Quantity:
+            self._pressure = pressure.to_base_units()
+        else:
+             self._pressure = pressure * u.Pa
```

ngam:
> * A UnitStrippedWarning pops up when running `particle_test`. Need help diagnosing the issue

This issue is not new, so it has nothing to do with your additions here. I do agree we need to look into it; it could be simply that `pint` doesn't play well with `numpy` in this particular case? There are a few upstream issues. Maybe worth opening this in a separate issue imo.

wkchuang:
> > > How will the user give us units? They will need access to `u`, right? They can either set their own `u` (i.e. `import particula; import pint; from pint import UnitRegistry; u = UnitRegistry()`) or they will have to use our `u` (i.e. `from particula import u` or `import particula; u = particula.u`). Can we make sure we are expecting the same `u` from the user as our package is defining?
> > > Additionally, what if the user happens to import `pint` for a separate package/project they're using in the same environment?
> > 
> > 
> > If I'm understanding the premise correctly, then when users create their own `u` or import pint separately, those registries would clash with our `u` and throw an error if they ever meet. If they're building on our work, they should quickly clash. We can leave clear instructions on following our units process and a warning on the error.
> 
> Yes, but that would introduce a lot of potential restrictions (e.g. what if the user wants to use `pint` separately or with another package?). An easier solution is to just assume everything comes in standard units from the user. I personally view units as more of a testing mechanism rather than a full-fledged user-facing mechanism. Who's in our audience here? People interested in using the underlying calculations or people who are interested in a super-tight unit system?

I think those potential restrictions apply regardless of how we implement units. If they import pint separately, then that comes with separate registries and they'll figure out a way to blend those with ours, regardless of whether we use pint or create our own pint backend. I think it's beyond our current scope and is left for a future developer who wants to use other packages. That might be us, but we cross that bridge when we come to it.

Why not both? The first audience is interested in the model, the second audience is interested in the usability, and I think they aren't mutually exclusive. In our world where we use plenty of other units, we could make it flexible for users to input nanometers, square centimeter per cubic meter, particle # per cc, atmospheres, celsius, whatever units their instruments output or papers use.  But, if we want to make standard unit assumptions, then we'd better make it very, very clear the units the code expects as the user is inputting values, because the code will take its own units and run with it.

particula/__init__.py
wkchuang: That's a good question, and I'm investigating further. Pint defines kelvin as degK, I will dig further to see if they are equivalent.

wkchuang:
> > I'll see if I can define a test. It failed when I tried running an input without units, I just have to figure out how to catch that and prevent the whole code from stopping.
> 
> There should be `assert == fail` type of thing in pytest. Try this: `with pytest.raises(ValueError)` from [here](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions) (note the unit error is going to be a value error, I think)

Thanks for the link!

ngam:
> Edit: I generally want the user to be aware of what they are inputting and what the code will do to the inputs. But it's fine if we end up going with the assumption and highlight that clearly

Yes, I agree with you. I am worried that it becomes burdensome to input the units, so a shortcut is to take "unitless" inputs  as standard units, while also allowing users to input some units (for now they must be from our `u` or they will fail --- later we can work on generalizing that: We can strip each input into magnitude and unit and then convert to our `u` regardless to where they come from). See my suggestions above for a possible implementation:

```diff
-    @u.wraps(None, [None, u.degK, u.Pa])
     def __init__(self, temperature, pressure):
-        """Function calls for enviornment class."""
+        """ Function calls for environment class.
+
+            Checks if the user provides a unit; if so, take to base units.
+            If user doesn't provide a unit, assume base units.
+        """
-        self._temperature = temperature
-        self._pressure = pressure
+        if type(temperature) == u.Quantity:
+            self._temperature = temperature.to_base_units()
+        else:
+            self._temperature = temperature * u.degK
+        if type(pressure) == u.Quantity:
+            self._pressure = pressure.to_base_units()
+        else:
+            self._pressure = pressure * u.Pa
```

particula/__init__.py
ngam: Thank you for looking in these things! I tried figuring this out and I am still a bit confused. Note `u.K` worked fine previously, so they could be interchangeable, but I am not sure.

particula/__init__.py
```diff
@@ -1,4 +1,4 @@
-""" a simple, fast, and powerful particle simulator.
+""" A simple, fast, and powerful particle simulator.
```
ngam, line 1: Btw, these lowercase edits were inspire by the following:

```
>>> import particula
>>> help(particula)
Help on package particula:

NAME
    particula - a simple, fast, and powerful particle simulator.

DESCRIPTION
    particula is a simple, fast, and powerful particle simulator,
    or at least two of the three, we hope. It is a simple particle
    system that is designed to be easy to use and easy to extend.
    The goal is to have a robust aerosol (gas phase + particle phase)
    simulation system that can be used to answer scientific questions
    that arise for experiments and research discussions.
    
    The main features of particula are:
    ...
    
    More details to follow.

PACKAGE CONTENTS
    aerosol_dynamics (package)

DATA
    u = <pint.registry.UnitRegistry object>

VERSION
    0.0.3

FILE
    .../python3.9/site-packages/particula/__init__.py

(END)
```

In particular, see

```
NAME
    particula - a simple, fast, and powerful particle simulator.
```

So this line often appears after the name of the file (module, etc.) and a dash. I thought it would look nicer if they were all lowercase to start. But this is just being silly on my part. Also, the little space between the `"""` and the top line is to make sure indentation is the same (`"""` = 3 spaces, and our Python indentation is 4 spaces)

Eventually, I want to figure out a way to automate making these docstrings and comments into actual documentation in [here](https://uncscode.github.io/particula/documentation/d000_detailed_documentation.html)

wkchuang: Thanks for explaining that! It makes way more sense now that it is lower case because it comes after a dash. I can revert those changes.

particula/__init__.py
wkchuang: After testing, `u.K` and `u.degK` are equivalent, and they interact normally together. I imagine both were implemented because `u.K` is the more accurate definition, and `u.degK` is for clarity and uniformity with `u.degF` and `u.degC`. I am indifferent between the two, but we should try to stick with one. Any opinions?

ngam:
@wkchuang, could you please close this now that we are doing a restructuring, it will likely be moot? Thanks! Please leave the associated issue open for now for visibility.

PR #85, opened 2022-01-01
correctly set branch for colab, binder, etc.

### LINKED ISSUE #83: figure out a way to make binder/livecode run the latest tag
https://github.com/uncscode/particula/blob/6ecb6a1a3a4b8e56c1a5b0e1f94f38638c64ca86/docs/_config.yml#L16

instead of the latest updates in main.

We could also isolate main --- i.e. introduce a new dev or work branch.

ngam:
Fixes #83 

This sets the branch to v0.0.3 (for now, but will go up later) instead of main. This way, the documentation are run for the latest release rather than main (which often contains newer additions after releases).

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -13,7 +13,7 @@ launch_buttons:
 repository:
   url: https://github.com/uncscode/particula
   path_to_book: docs
-  branch: tags/v0.0.3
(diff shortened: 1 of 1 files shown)


Gorkowski (approved):
ok, with me :)

PR #87, opened 2022-01-01
Fixing UnitStrippedWarning

### LINKED ISSUE #86: > * A UnitStrippedWarning pops up when running `particle_test`.
> * A UnitStrippedWarning pops up when running `particle_test`. Need help diagnosing the issue

This issue is not new, so it has nothing to do with your additions here. I do agree we need to look into it; it could be simply that `pint` doesn't play well with `numpy` in this particular case? There are a few upstream issues. Maybe worth opening this in a separate issue imo.

_Originally posted by @ngam in https://github.com/uncscode/particula/issues/84#issuecomment-1003608779_

ngam:
Fixes #86

Goes directly into branch wkchuang/issue82 and should be merged before #84

### DIFF
--- a/particula/aerosol_dynamics/tests/particle_test.py
+++ b/particula/aerosol_dynamics/tests/particle_test.py
@@ -146,7 +146,7 @@ def test_dimensioned_coagulation_kernel():
 
     assert small_particle.dimensioned_coagulation_kernel(
         large_particle, standard_environment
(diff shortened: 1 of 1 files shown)


ngam:
@wkchuang This resolves the issue with that numpy warning. Please merge this into your branch, so that this issue is resolved. (The solution is very minor)

PR #91, opened 2022-01-17
added particle coagulation in parcel class

### LINKED ISSUE #47: add particle to particle interacations in parcel class
https://github.com/uncscode/particula/blob/85ba3f80b2f6d5b7d87c0e9f1a848d1bb9d86e94/aerosol_dynamics/parcel.py#L98

Gorkowski:
Fixes #47

### DIFF
--- a/particula/aerosol_dynamics/parcel.py
+++ b/particula/aerosol_dynamics/parcel.py
@@ -4,7 +4,6 @@
 import numpy as np
 from particula.aerosol_dynamics import particle
 
-
 class Parcel:
     """
     Creates an air parcel to track and interact with a list of particle
@@ -29,6 +28,7 @@ class Parcel:
         self._name = name
         self._particle_data = np.array([])
         self._enviroment = parcel_environment
+        self.unit_warning_flag = True
 
     def name(self) -> str:
         """Returns the name of the parcel."""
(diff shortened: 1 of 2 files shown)


particula/aerosol_dynamics/parcel.py
```diff
@@ -1,21 +1,20 @@
-"""class for creating a air parcel to put particles and gases in.
+"""class for creating an air parcel to put particles and gases into.
 """
 
 import numpy as np
-from particula import u
```
ngam, line 5: Can you bring this back and do some unit checks as decorators?

ngam (changes requested):

particula/aerosol_dynamics/parcel.py
Gorkowski: We can talk about how we want to handle units, based on @wkchuang pull draft. I ran into unit errors/mismatches when assigning units in the inputs (so I removed them). I think it came from reloading the module and having inconsistent instances of unitreg, but not completely sure.

Gorkowski: I added back the unit check for radius. The mismatch was due to multiple instances of unitreg during modifying parcel.py. 
We do not need 'from particula import u' in as either the units will be in the input or the particle class will assign them, no units get created in parcel class. We don't need 'u' to do the unit checks, with the current try-exempt check being used.

ngam (approved):

particula/aerosol_dynamics/parcel.py
ngam: Maybe we should consider taking all units out and place them in the test suite. I think that makes significantly more sense to me than having units around in the actual routines. We will keep talking about this in PR #84

PR #93, opened 2022-01-20
How does the parcel class work

### LINKED ISSUE #92: How does the parcel class work
Can we get a basic notebook on how the parcel class works?

Gorkowski:
Fixes #92

Jupyter  notebook on the basics of the parcel class

### DIFF
--- a/docs/tutorials/002_Parcel_Class.ipynb
+++ b/docs/tutorials/002_Parcel_Class.ipynb
@@ -1,0 +1,549 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Many particles\n",
+    "Next well be moving through the code to see how to create a `parcel` object that will be used to track multiple particles. First we will need to import the necessary modules."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "os.chdir('..') # move up two levels for local testing, otherwise comment out if you have installed the package\n",
+    "os.chdir('..')\n",
+    "\n",
+    "from particula import u\n",
+    "from particula.aerosol_dynamics import environment, particle, parcel\n",
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


docs/tutorials/002_Parcel_Class.ipynb
```diff
@@ -0,0 +1,549 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Many particles\n",
+    "Next well be moving through the code to see how to create a `parcel` object that will be used to track multiple particles. First we will need to import the necessary modules."
```
wkchuang, line 8: `well` to `we'll`

docs/tutorials/002_Parcel_Class.ipynb
```diff
@@ -0,0 +1,549 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "DimensionalityError",
+     "evalue": "Cannot convert from 'nanometer ** 3' ([length] ** 3) to 'kilogram' ([mass])",
```
wkchuang, line 457: Solved. Within `particle.py`, the units enforcement/assignment for radius and density occur when those methods are called. So even though `missing_units_parcel` was given no units for its density values, calling the `density` method will assign and output the default units. 

The problem arises when the mass assignment happens using the unitless inputs. Here `density`, taken straight from the input, has no units, and causes an error. 
`self._mass = density() * (4*np.pi/3) * (radius()**3)`

If instead we take it from the method calls `self.density()`, then it will calculate mass with no issues:
`self._mass = self.density() * (4*np.pi/3) * (self.radius()**3)`

ngam:
Hey, sorry for not engaging with this earlier. It's hard to reviewer notebooks this way. I recommend we move this to Google colab. I send you a link momentarily.

ngam:
@Gorkowski @wkchuang check this out, you can even comment on this :)

https://colab.research.google.com/drive/1nP1bcMRL4rP1nvouEFD4rkccehM96-NY?usp=sharing

ngam:
Generally, I am okay with a loose review process for these notebooks going forward. I don't want to burden us with this as heavily, so long as we are writing enough documentation in docstrings and so on. We can always have a documentation "week" after a release.

ngam:
We will always rerun (re-execute) the notebooks upon release anyway.

Issue #73, opened 2021-12-25
add cli features


ngam:
**Is your feature request related to a problem? Please describe.**
It would be helpful to have some basic command-line features that allow quick access to some utility methods.

**Describe the solution you'd like**
Add console scripts with entry points. This allows the execution of a command directly without needing to invoke Python. As an example, to return the binary coagulation rate coefficient,

```bash
$ particula binary_coagulation 1e-9 2e-9
XXX.YYYeZZZ cc/s
```

**Describe alternatives you've considered**
Cleary outline how to invoke modules from the command line, e.g. `python -m xyz` or `python -c "xyz"` but both are a bit more involved.

**Additional context**
Focus on the convenience and quick utility/calculation in the midst of something else.

ngam:
Keep

PR #99, opened 2022-02-06
WIP: Reorganize

ngam:
- continuing @Gorkowski's branch


aktwo:
Some high level thoughts:
- First of all, thanks for thinking about the high level organization and trying to move it forward :)
- It looks like this PR is 2K+ lines of diffs. Reviewing that much code thoroughly is very difficult, so in the future I'd suggest keeping pull requests to 100-200 lines tops. There are many articles out there on the value of small pull requests, here's one: https://essenceofcode.com/2019/10/29/the-art-of-small-pull-requests/
- Given that this PR is so large (and sets a lot of the foundation for the architecture of the project), I'd suggest that we first take a step back and discuss the underlying principles/vision that you had in mind for this change (which I propose we do on this thread for now). Once we get consensus here, let's write it down in the README (or another such file) so we can refer back to it and put any code changes into context.
- I'd suggest we split this chain of commits up into smaller PRs, where each PR is its own logical change. It will also be useful to add a quick high level explanation for each change and the motivation behind it. That might be a good starting point for us to discuss the high level organization (as well as good practice for writing small PRs in the future!).
- I realize this will all take more time than merging 2K lines of code in a single PR, but I think it's worth our time to be thoughtful about this structure, because once it's set it'll be significantly harder to change later. It's also a great opportunity for us to practice collaborating as software engineers :)

I'd suggest that the next step is for @ngam or @Gorkowski to share their organization thought process in this thread. Just my $0.02, would love to hear what everyone else thinks!

ngam:
> * I'd suggest we split this chain of commits up into smaller PRs

Yes, I think we can do this without much additional work.

For the most part, this PR does NOT add anything substantial to the work that existed before it and it won't break anything that worked before. We actually met a few weeks ago and we found ourselves hitting a wall --- and we discussed a few strategies (including doing away with units), but after reflection, I personally thought we could _reorganize_ the whole thing without any breaking changes.

The simple idea is that anything that existed in the classes we had (e.g. Particle) should be refactored away and modularized. For example, we have a method `knudsen_number` in class `Particle` in file particle.py --- so we can call it with `from particula.aerosol_dynamics.particle import Particle; kn = Particle(...).knudsen_number()` which takes us where the Knudsen number is calculated. Previously we had those lines of code in `particula.aerosol_dynamics.particle.py`. Now, we will have those lines of code in `paritcula.utils.particle_.calc_knudsen_number` under a function called `knudsen_number`. The premise is that these utils are hidden away and organized so that they are portable and editable with ease (each file will end up much simpler and relatively lightweight, with space to add thorough documentation (and testing) about what this little utility does. However, in place of the lines of code in `particula.aerosol_dynamics.particle.py`, the function is still called all the same and should yield exactly the same result. Hence, `from particula.aerosol_dynamics.particle import Particle; kn = Particle(...).knudsen_number()` will still work, but now we have the actual underlying code stored elsewhere and it is isolated and modularized so that it could be used in different classes and it no longer has to be extracted from the `Particle` class. 

I am more than happy to start doing smaller chunks in PRs soon. So that it will be like: "Here, we refactor x method and we improve its testing this way way" --- we will have ~10 lightweight PRs that are much more manageable (involving essentially 3--4 edited files with some minor edits in other files (e.g. `__init__.py`).

ngam:
Previously:

https://github.com/uncscode/particula/blob/719080530e79342c7842da56c3d496b573d4b342/particula/aerosol_dynamics/particle.py#L109-L121

with tests:

https://github.com/uncscode/particula/blob/719080530e79342c7842da56c3d496b573d4b342/particula/aerosol_dynamics/tests/particle_test.py#L70-L92

---

Now:

https://github.com/uncscode/particula/blob/4cf6e97d39b60e399c8df8694d189c7faaecfb3b/particula/utils/particle_/calc_knudsen_number.py#L7-L39

with tests:

https://github.com/uncscode/particula/blob/4cf6e97d39b60e399c8df8694d189c7faaecfb3b/particula/utils/particle_/tests/knudsen_number_test.py#L1-L35

and tests:

https://github.com/uncscode/particula/blob/4cf6e97d39b60e399c8df8694d189c7faaecfb3b/particula/aerosol_dynamics/tests/particle_test.py#L55-L62

ngam:
Another "change" is enabling defaults --- this is not really a change, it just allows more flexibility with no providing parameters repeatedly (e.g. temperature, density, etc. which often don't deviate much and don't matter as much for quick calculations when compared with size effects)

ngam:
Let me know if you'd like me to start splitting the work into smaller PRs targeted one utility at a time. There will be around ~10 of them. I can start sending those PRs tomorrow :) 

@Gorkowski @wkchuang @aktwo --- please vote 🗳️

ngam:
> * I'd suggest we split this chain of commits up into smaller PRs, where each PR is its own logical change. It will also be useful to add a quick high level explanation for each change and the motivation behind it. That might be a good starting point for us to discuss the high level organization (as well as good practice for writing small PRs in the future!).
> * I realize this will all take more time than merging 2K lines of code in a single PR, but I think it's worth our time to be thoughtful about this structure, because once it's set it'll be significantly harder to change later. It's also a great opportunity for us to practice collaborating as software engineers :)

Strongly agree with these two points! I just couldn't figure out a way to split things without being pushed to do so 😆

ngam:
> Sounds good to me!

Okay, I will start sending PRs this PM.

ngam:
I added PRs:
#107 to outline the proposed structure
#108 and #109 to show it in practice.

I will pause to get feedback from you before proceeding. I can file all the remaining PRs shortly, but it would be good to get feedback about naming, strategies, etc. at an earlier stage 👍 

Thanks team ❤️

aktwo:
Thanks for the context! 

> The simple idea is that anything that existed in the classes we had (e.g. Particle) should be refactored away and modularized.

For my understanding, what problem are we trying to solve by doing this? You mentioned hitting a wall with something. My mental model is that the current methods in the particle class primarily make sense in the context of particles (i.e. Knudsen's number), but I'm not as familiar with the science as everyone else here is :)

> but now we have the actual underlying code stored elsewhere and it is isolated and modularized so that it could be used in different classes

Do you have an example of another class that we might want to call these methods from?

> with space to add thorough documentation (and testing) about what this little utility does). 

Sorry if I'm missing something, but I'm not sure I see what about the separate utility functions enables documentation/testing that isn't already possible with instance methods?

> Another "change" is enabling defaults

At a high level this seems like a great idea! I'd be happy to approve a pull request that sets sane defaults for parameters that we don't expect to need to tweak that often.

ngam:
The current setup with the Particle class is difficult to scale up for many many particles (imo). We are talking 1e5 particles per cc (centimeter cubic). @Gorkowski is actively researching so-called parcel physics (for his real day job), which could be thought of as a collection of particles interacting with the environment surrounding them (and we want them to interact amongst themselves).

We haven't yet settled on the way forward --- but there are generally speaking two distinct ways to tackle this in my mind.

1. the classic way (sectional bins): This could be thought of as simple the Particle class we have + a few new attributes. We can achieve this by simply defining a new class, call it `ParticleBin` that inherits from `Particle` and we add some many-particle attributes to it (something about the concentration (i.e. number and volume); something about the definition of the bins they occupy (i.e. not only a radius, but also radius bounds); and so on).
2. a "continuous" way (that's how I personally work on these problems): This could be achieved by introducing a new concept to replace Particle, let's call it `ParticleDistribution` where some specific properties of it are assigned (e.g. like a normal distribution with mean, standard deviation, etc. --- you can think of these as distribution moments, though that's what people often call the first point above so let's not confuse them). In ParticleDistribution, we sort of lose touch with particles as discrete entities and some adjustments will have to e made --- those adjustments in reality, are going to be simple tweaks in these methods (like Knudsen number, etc.) to allow them to accept either the defining properties of a distribution or a sectionalized version of it (arrays). 

For 1, we can easily just keep stacking the Particle class ad infinitum, but that's bound to get hard (at least for us scientists) --- again, we are talking about something on the order of millions. This is not only hard to code and keep track of, it may prove to be computationally difficult (coagulation is definitely the rate-limiting step in all of this btw). However, I can see us adopting 1 and seeing how that goes. However, by splitting the methods and refactoring them away, we are able to pursue both 1 and 2 simultaneously and switching back and forth.

Another issue is carrying the units everywhere we go --- which is new to us, and we actually love it, but it gets tricky. The Particle object becomes overwhelming...

Other classes will emerge as we go on, I think. Yes, these are particle-specific methods, but they are also universal to an extent. As a superficial example, again with Knudsen number: Knudsen number is really just a measure of the deviation from continuum effects (it is 0 in the continuum regime where particles are larger; very large in the kinetic regime where particles are smaller). Note how "particle" defines what Knudsen number is, right? However, these dimensionless numbers (Knudsen number is dimensionless) are often there for a something more nuanced and more fundamental --- so there will be situations where you just need the Knudsen number and nothing else in the Particle class is relevant to you (i.e. you only care about the size of the particle, and the mean free path of the medium it is in); using just that you can derive other values and pursue other ideas. How's this helpful to anyone? Well, I'd would like to convert some of these methods into command line utilities at some point (#73). The motivation here is, we have the scripts and methods already, so why not expose them so that someone like me can just type `particula knudsen radius=1e-9` to get a value (it should be around 65). 

>> with space to add thorough documentation (and testing) about what this little utility does)
>
> Sorry if I'm missing something, but I'm not sure I see what about the separate utility functions enables documentation/testing that isn't already possible with instance methods?

It's possible to do it all in one script. My point here is that it will be _easier_ to handle the testing and documentation of these functions if they are isolated elsewhere (e.g. the Knudsen number has its own file.py and file_test.py). Once the refactoring and modularization are done, we are going to immediately ensure that each method can take arrays as inputs --- that's distinctly difficult to achieve if they are inside the Particle class because (for now) that's written for only one particle (in my understanding).

Regarding the defaults: There are a lot of details in our coagulation calculation that don't really matter for most cases (i.e. they could be much more easily calculated using simpler techniques) and that's because a lot of these optional variables have little effect in many cases, but are really important in special cases. So yes, for me personally, the rule of thumb about particles is their size --- almost everything else about them is a higher-order dependency (i.e. less important).

aktwo:
Thanks for the context! I agree that for 1e5 or more particles, an object-per-particle might be a little cumbersome, even with fancy optimizations like using `__slots__` to save memory.

I totally understand why splitting these functions out into their own files might make sense. Before we commit to this particular structure though, what do y'all think about sketching out some high level pseudocode of the kinds of operations we want to eventually do? That might give us a way to sanity check that the abstractions we're building are going to empower us to do the operations we're interested in. The main reason I'm having difficulty reviewing all the new PRs (thanks for splitting them up btw 🙏) is because as a non scientist I don't have a good idea of what the desired end state is.

Definitely not trying to slow you guys down though, so let me take a pass and see if I have any significant high level comments in parallel (will probably be tomorrow because I'm heading to bed soon).

ngam:
No problem @aktwo, please take your time. I am sorry if this seems like pressure-inducing! I (and I am sure @Gorkowski and @wkchuang would echo the same sentiment) appreciate your feedback --- by slowing us down, you make us to reflect and assess our choices and that's very helpful.

First and foremost for me is that everyone is on the same page (roughly speaking, without necessarily knowing all the details) so that we are not leaving anyone behind. I wouldn't worry about fully consolidating our efforts just yet --- and it would be okay for every-/any-one to pursue tangential ideas/interests. The deeper motivation behind this restructuring is NOT changing our direction per se, but expanding it and allowing greater flexibility. For the time being, I don't see anything breaking or changing. We will probably push 0.1.0 once that's the case (once we have a many-particle coagulation calculation, that would be a good 0.1.0 release)

I will try to slim down the PRs further going forward --- please note, they should be thought of as a sequence: #107, then #108 and finally #109 (they build on top of each other). I thought that would give people a better idea how this proposal would work in practice... but maybe I was wrong 🌗

PR #102, opened 2022-02-06
Gorkowski/issue95

ngam:
creating @Gorkowski PR here for visibility


aktwo:
It looks like this is a subset of https://github.com/uncscode/particula/pull/99. If so, consider dropping this PR so that we can consolidate discussion there?

ngam:
> It looks like this is a subset of #99. If so, consider dropping this PR so that we can consolidate discussion there?

Hi @aktwo, yes all current 4 PRs are messy and will all coalesce around one soon. This is because we are trying to modularize the code better by refactoring all methods into separate utilities (particula/utils/). Hopefully we will be able to review and merge tomorrow or the day after :)

PR #107, opened 2022-02-09
proposal to restructure methods

### LINKED ISSUE #103: outline restructuring
- why do this now
- how it is being done
- roadmap for next PRs

ngam:
Fixes #103 

Highlights:
- add utils folder with subfolders
- proposing `_` prefix for these utils folders
- changed version from 0.0.4-dev0 to 0.0.4.dev0 to follow proper sem ver guidelines

Notes:
- since a lot of the work has been done, more PRs can simply build on this and be submitted before approval and merge here

Motivation:
- a lot of these methods should be verified and tested individually and not within a class; they are often universal.
- methods should be exposed and flexible to wide array of inputs and outputs desired:
   - including unit-endowed and unitless inputs
   - including array-like or single-valued inputs
- generally easier to isolate errors and edit to improve and sustainably maintain

### DIFF
--- a/particula/__init__.py
+++ b/particula/__init__.py
@@ -19,4 +19,4 @@ from pint import UnitRegistry
 # u is the unit registry name.
 u = UnitRegistry()
 
-__version__ = '0.0.4-dev0'
+__version__ = '0.0.4.dev0'
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ jupyterlab
 pytype
 typing
 build
+flake8
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,6 +38,7 @@ dev =
     pytype
     typing
     build
+    flake8
 
 [options.packages.find]
 where = .


ngam:
Since this is a restructuring, I would like at least *two* approvals from @aktwo, @wkchuang, and @Gorkowski before merging.

Gorkowski:
I don't know which prefix/suffix is better, and do not have a preference. I do like having somthing there, to show that this is different from the main functions classes/methods that most people would use

Gorkowski (approved):

wkchuang:
I think this is a case where @aktwo is best suited to answer. I'm not sure if there are any considerations we should be aware of. If not, I might prefer before more than after (it's more clear that this is different). An underscore afterwards is harder to notice

ngam:
> My preference is to not use underscores in any folder names, but use a leading underscore for any functions that we want to be private as per the style guide IIUC.
> 
> The main reason is that I could see us potentially wanting to having public utility functions in the future, in which case we'd have two parallel file trees: `_utils/environment` and `utils/environment` which feels cumbersome.

I am fine with avoiding underscores in file/folder names

aktwo:
Cool! It looks like common/environment/particle might still have underscores? I'm on mobile though so maybe I'm missing something

ngam:
I haven't made the edits yet

aktwo:
Understood. Another nit: let's omit the empty __init__.py files until we need them?

aktwo:
I looked at #108 and #109 and had another thought: for simplicity, wdyt about reducing folder nesting, so something like:
- particula/constants.py
- particula/tests/test_constants.py
- particula/units.py
- particula/tests/test_units.py
- particula/environment.py
- particula/tests/test_environment.py

The general principle here is to avoid adding additional folder structure until we've written more code and see the need for it. Once things start to feel cluttered, we can always use automated refactoring tools (e.g. https://realpython.com/python-refactoring/#using-rope-for-refactoring) to do all the grungy work for us. My motivation for this is to prevent us from creating catch-all folders like "utils" or "common", because I've seen them become difficult to make sense of as projects grow larger. Instead, I'd prefer folders with names that are well-scoped and thematic, so it's super clear what the criteria for membership is.

Lmk what you all think!

ngam:
> I looked at #108 and #109 and had another thought: for simplicity, wdyt about reducing folder nesting, so something like:
> 
> * particula/constants.py
> * particula/tests/test_constants.py
> * particula/units.py
> * particula/tests/test_units.py
> * particula/environment.py
> * particula/tests/test_environment.py
> 
> The general principle here is to avoid adding additional folder structure until we've written more code and see the need for it. Once things start to feel cluttered, we can always use automated refactoring tools (e.g. https://realpython.com/python-refactoring/#using-rope-for-refactoring) to do all the grungy work for us. My motivation for this is to prevent us from creating catch-all folders like "utils" or "common", because I've seen them become difficult to make sense of as projects grow larger. Instead, I'd prefer folders with names that are well-scoped and thematic, so it's super clear what the criteria for membership is.
> 
> Lmk what you all think!

This will work well for environment and "common" utils for now, but it will become cluttered once we add the particle ones.

I do like eliminating `aerosol_dynamics` for now though... 

Also, just to be clear: I want to use the folder structure to only organize stuff for us (developers), but once ready, all these methods will be exposed through `__init__.py` files. How that will work in practice is TBD (I don't know). However, I am fine with following your approach above for now (and we can adjust and tinker with the paths as time goes on --- it is pretty easy to fake where things are in terms of source code, we can alway sym-link things through `__init__.py`).

ngam:
I will arrange all these edits and submit them later tonight or earlier tomorrow.

ngam:
But to be clear, I don't intend to get rid of `aerosol_dynamics` yet. I just want to isolate these methods, but leave everything working as is for the time being. We make _breaking changes_ once we are done with the restructuring.

ngam:
> we can always use automated refactoring tools (e.g. https://realpython.com/python-refactoring/#using-rope-for-refactoring) to do all the grungy work for us.

Some of these utilities are already built-in VSCode btw ;)

PR #108, opened 2022-02-09
setting up constants in utils._common

### LINKED ISSUE #106: set up constants and expose them in utils
- consider adding a yaml parser
- expose in `__init__.py` for portability

ngam:
Fixes #106 

Builds on #107 (Closes #107 or should be merged after it)

Highlights:
- added constants with units in utils in a more centralized location
- added simple tests

TODO:
- allow override with user-supplied configuration file in the future

### DIFF
--- a/particula/__init__.py
+++ b/particula/__init__.py
@@ -19,4 +19,4 @@ from pint import UnitRegistry
 # u is the unit registry name.
 u = UnitRegistry()
 
-__version__ = '0.0.4-dev0'
+__version__ = '0.0.4.dev0'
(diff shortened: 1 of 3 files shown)


Gorkowski:
do you mean an override to change what constants to use? 

the rest looks good.

Gorkowski (approved):

ngam:
> do you mean an override to change what constants to use?

Yes, or if someone wants to have a slightly different constant (instead of 5.3673 they want 5.367398332 for example). We are using really high-quality constants though, but it is good to give people an option to tinker with things to see how things will change. Also, these are good options to have for sensitivity studies, if someone wants to ask, "How sensitive is the coagulation kernel to the elementary charge value?"

wkchuang (approved):

PR #120, opened 2022-02-16
Following up on PR 109

ngam:
- Minor edits to PR #109 
- Assigning to @aktwo


particula/constants.py
```diff
@@ -1,47 +1,35 @@
...
 RELATIVE_PERMITTIVITY_AIR = 1.0005  # unitless
 
-# permittivity of free space in F/m
+# Permittivity of free space in F/m
 # Also known as the electric constant, permittivity of free space
 # Often denoted by epsilon_0
 VACUUM_PERMITTIVITY = 8.85418782e-12 * u.F / u.m
 
-# permittivity of air
 ELECTRIC_PERMITTIVITY = RELATIVE_PERMITTIVITY_AIR * VACUUM_PERMITTIVITY
```
aktwo, line 30: Do we want a comment for the units to be consistent with the gas_constant above?

particula/constants.py
```diff
@@ -1,47 +1,35 @@
...
 
-# viscosity of air at 273.15 K
-# these values are used to calculate the dynamic viscosity of air
+# These values are used to calculate the dynamic viscosity of air
```
aktwo, line 32: Doesn't need to be in this PR, but it seems like these values are dependent on temperature, so we may want to move them into a separate file!

particula/tests/unitless_test.py
```diff
@@ -1,24 +1,22 @@
-""" A quick test of strip_units.unitless utility
```
aktwo, line 1: Nit: rename the file to units_test to be consistent with the source file?

aktwo (approved):
Thanks for following up! I left a few minor comments but generally LGTM :)

particula/constants.py
ngam: No no, these values are constant. Look at my other PR to get a more thorough understanding of how this works in practice, #122 or see this link which I am going to cite in the updated docstring           there https://resources.wolframcloud.com/FormulaRepository/resources/Sutherlands-Formula. The *resulting* dynamic viscosity is the one that's changing; I know, I know, such a misleading name to call it "dynamic" but oh well... 

Let me take this opportunity to actually re-answer one of your earlier questions about the need to isolate these methods. Essentially, this actually goes to illustrate why it is best to have these as isolated methods. Why? Because this particular formula actually applied to any _ideal_ gas --- and we are assuming an ideal gas behavior of air here. So what happens if we want to calculate this in a different gas (unlikely) or if we want to relax the ideal gas assumption? Having these utilities be independent and isolated will make it much easier to tweak them without having to tweak the underlying code for the calculating the end products. For example, in the following PR (#122), I add a lot of options for the user to tweak this, but I haven't added the deviation from the ideal gas yet (I actually don't know how to do it because it is so niche and unlikely to change things --- but that's where science actually happens, at the edge cases). Anyway, so hopefully this starts to make more physics sense to you.

particula/tests/unitless_test.py
ngam: Yes, good catch!

particula/constants.py
ngam: It's implied since `RELATIVE_PERMITTIVITY_AIR` is unitless, so `ELECTRIC_PERMITTIVITY` will carry the units of `VACUUM_PERMITTIVITY`.

particula/constants.py
ngam: Oh! Maybe your point is that they're linked to temperature in a way? Yes, but they're anchored at a temperature, hence `REF` in the name. So, these are considered constants imo, but we may want to change the names slightly to reflect the specificity

ngam:
Merging this after approval, follow along in PR #122

PR #122, opened 2022-02-17
add dynamic viscosity utility

ngam:
- generalizing and extending the dynamic viscosity calculation
- adding user warning + control flow around inputs
- some generic tests (suggestions are welcome)

Paging @wkchuang for review, thanks!

Notes: 
- do you prefer `dynamic_viscosity` or `dviscosity`? We really only say `dynamic` to differentiate it from `kinematic`, so it should be pretty obvious `dvis...` and `kvis...` would refer to the dynamic and kinematic viscosities respectively...  
- this takes scalar, list, and numpy.array as input just fine --- returns numpy.array in the case of list and numpy.array. Should add a test later to formalize this later.


ngam:
> Why is this in a new folder /util and not in /utils ?

Indecisive lol... do you have a preference?

PR #123, opened 2022-02-18
starting a list of spellcheck overrides in devcontainer 

ngam:
- This is a simple addition to the devcontainer to override the spellcheck. Usually I would add these types of spell checks override to my personal user account, but I thought sharing them here may make using the docker container more useful/easier. Please feel free to add more to this list as you wish.
- Also decided to move `pip install -e .[dev]` to the Dockerfile because that way it won't happen every single time upon reopening the container


Gorkowski:
Send it :)

Gorkowski (approved):

PR #124, opened 2022-02-18
modularizing mean free path calc

ngam:
This should be reviewed after #122 since it depends on it. Only mean_free_path.py and mean_free_path_test.py are relevant for this PR. Everything else should be reviewed in the other PR. If you review before #122 is merged, then please approve and I can merge it after the other one is merged.

Assigning to @Gorkowski.

Highlights:
- added a calculation for mfp + test
- generalized from before to allow input of different molec_wt

Notes:
- currently this takes inputs that are list or np.array, it returns np.array, so it accepts 300, [300, 305], and numpy.array([300, 305]) as inputs. Should add a test and confirmation of this behavior later. This is important when doing a varying temperature (ascending parcel, etc.)


ngam:
How do you feel about this, @Gorkowski?

Gorkowski (approved):

PR #125, opened 2022-02-19
adding particle mass and reduced mass utils

ngam:
- simple mass calc, ensuring arrays work + test
- generic reduced_quantity function + test

### DIFF
--- a/particula/util/reduced_quantity.py
+++ b/particula/util/reduced_quantity.py
@@ -1,0 +1,81 @@
+""" calculating reduced quantity.
+
+    reduced_quantity =
+        quantity_1 * quantity_2 / (quantity_1 + quantity_2)
+"""
+
+import numpy as np
+from particula import u
+
+
+def reduced_quantity(a_quantity, b_quantity) -> float:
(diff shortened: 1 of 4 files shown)


particula/util/reduced_quantity.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+        if a_q.to_base_units().units != b_q.units:
+            raise TypeError(
+                f"\n\t"
+                f"{a_q} and {b_q} not compatible!\n\t"
+                f"Quantities must have same units to be reduced.\n\t"
+                f"Try: {b_q} and {a_q} {b_q.units} for example"
+            )
+
+    if isinstance(a_q, list):
```
Gorkowski, line 76: cool, so it doesn't matter which input is an array

Gorkowski (approved):

particula/util/reduced_quantity.py
ngam: yes, but the problem here is more like you cannot divide/multiply lists... the interesting thing is that pint converts things to numpy.array anyway.

Issue #110, opened 2022-02-10
utilize *args and **kwargs in setting defaults


ngam:
It would be more meaningful and less laborious if we could boil down the essential defaults into `*args` and `**kwargs`. 

a reference/intro: https://realpython.com/python-kwargs-and-args/

Ideas: 
- use both, especially **kwargs
- define at the outset which parameters would fall under which category
- assign precedence (e.g. one could provide both mfp and temperature, but mfp is calculated from temperature, so which one should take precedence?)

PR #127, opened 2022-02-19
Particle distribution class using numpy arrays

Gorkowski:
My current attempt at an array for particles. Thoughts are welcome.

This includes basic particle number, mass, total mass, total number.
In addition, I added rasterization, which will be used in time-stepping so our number of particles doesn't increase uncontrollably.


ngam:
btw, see my `ffwd` branch: https://github.com/ngam/particula/tree/ffwd for the latest adventures. I am thinking of using kwargs and args like #110

ngam:
> In addition, I added rasterization, which will be used in time-stepping so our number of particles doesn't increase uncontrollably.

can you explain to me the motivation behind this? have you seen any issues so far?

Also, what is the meaning of concentration in your case?

I should have a very clean way to calculate numpy.array-based coagulation soon-ish, I just need to run through the code and add the tests and adjustments. See #131 #130

Gorkowski:
>>In addition, I added rasterization, which will be used in time-stepping so our number of particles doesn't increase uncontrollably.

>>can you explain to me the motivation behind this? have you seen any issues so far?
No issues so far, I just had it on my mind for two aspects. 1) for keeping the number of particles we need to track lower, but this may not be need. 2) saving out data in a pandas data frame with uniform size bins as we time-step the simulation. So depending how it gets used or doesn't we can move this method, to where it gets used most.

>>Also, what is the meaning of concentration in your case?
I was thinking total concentration. total number/m3.... total mass/m3.

ngam:
@Gorkowski, let's rebase this. Do you want me to try to do it or do you want to do it? I can probably do it quicker since I am familiar with the underlying code, but you will need to ensure I am not messing things up...

ngam:
Actually... we can achieve most of what this PR is doing through a modified particle class... with the new methods. The new methods take radius = 1, radius = [1], radius = [1, 2, 3, 4, ...], or radius = numpy.array([1, 2, 3, 4, ... ]) with or without units. What else does this PR need to accomplish (let's ignore the raster. stuff for now)?

Gorkowski:
Yeah you can rebase, I was going to do that in the morning anyway.

You can ignore raster.

I was going to use the new methods, and just call the coagulation. I didn't think the individual calls of everything else were necessary. 

If we can expand particle to do this, great. I just didn't know how much of a hassle that would be, it seemed to me that making a new class was clearer for me.

ngam:
> Yeah you can rebase, I was going to do that in the morning anyway.
> 
> You can ignore raster.
> 
> I was going to use the new methods, and just call the coagulation. I didn't think the individual calls of everything else were necessary.
> 
> If we can expand particle to do this, great. I just didn't know how much of a hassle that would be, it seemed to me that making a new class was clearer for me.

Okay, how about this, I will prepare a new class for you and share it in a new PR to illustrate how I think the new methods will work. Then you can take it from there and modify as you see fit. Hopefully soon, I will be able to add the height stuff and we can have a fully functioning suite to do real calculations. I also need to benchmark these calculations to make sure we are at least somewhat close to good values!

Gorkowski:
Yeah that is a good plan 🙂

ngam:
> and just call the coagulation

for this, you can simply try the following:

```
>>> from particula import u
>>> from particula.util.dimensionless_coagulation import full_coag
>>> full_coag(radius=1e-9)
<Quantity([[8.8332462e-16]], 'meter ** 3 / second')>
>>> full_coag(radius=[1e-9, 2e-9, 3e-9])
<Quantity([[8.83324620e-16 1.48330622e-15 2.51951248e-15]
 [1.48330622e-15 1.23535242e-15 1.54351423e-15]
 [2.51951248e-15 1.54351423e-15 1.49147865e-15]], 'meter ** 3 / second')>
>>> full_coag(radius=[1e-9, 2e-9, 3e-9], other_radius=[1e-9])
<Quantity([[8.83324620e-16]
 [1.48330622e-15]
 [2.51951248e-15]], 'meter ** 3 / second')>
>>> full_coag(radius=[1e-9, 2e-9, 3e-9], other_radius=[1e-9, 2e-9])
<Quantity([[8.83324620e-16 1.48330622e-15]
 [1.48330622e-15 1.23535242e-15]
 [2.51951248e-15 1.54351423e-15]], 'meter ** 3 / second')>
>>> full_coag(radius=[1e-9, 2e-9, 3e-9], other_radius=[1e-9, 2e-9], density=1e3, other_density=1e3, charge=0, other_charge=1, temperature=300, pressure=1e5)
<Quantity([[8.86389253e-16 1.48859282e-15]
 [1.48859282e-15 1.23990243e-15]
 [2.52872883e-15 1.54939954e-15]], 'meter ** 3 / second')>

```

Or slightly cleaner:

```
>>> from particula import u
>>> from particula.util.dimensionless_coagulation import full_coag
>>> full_coag(radius=1e-9).m_as(u.cm**3/u.s)
array([[8.8332462e-10]])
>>> full_coag(radius=[1e-9, 2e-9, 3e-9]).m_as(u.cm**3/u.s)
array([[8.83324620e-10, 1.48330622e-09, 2.51951248e-09],
       [1.48330622e-09, 1.23535242e-09, 1.54351423e-09],
       [2.51951248e-09, 1.54351423e-09, 1.49147865e-09]])
>>> full_coag(radius=[1e-9, 2e-9, 3e-9], other_radius=[1e-9]).m_as(u.cm**3/u.s)
array([[8.83324620e-10],
       [1.48330622e-09],
       [2.51951248e-09]])
>>> full_coag(radius=[1e-9, 2e-9, 3e-9], other_radius=[1e-9, 2e-9]).m_as(u.cm**3/u.s)
array([[8.83324620e-10, 1.48330622e-09],
       [1.48330622e-09, 1.23535242e-09],
       [2.51951248e-09, 1.54351423e-09]])
>>> full_coag(radius=[1e-9, 2e-9, 3e-9], other_radius=[1e-9, 2e-9], density=1e3, other_density=1e3, charge=0, other_charge=1, temperature=300, pressure=1e5).m_as(u.cm**3/u.s)
array([[8.86389253e-10, 1.48859282e-09],
       [1.48859282e-09, 1.23990243e-09],
       [2.52872883e-09, 1.54939954e-09]])
```

where `.m_as(u.cm**3/u.s)` is magnitude as cc/s. Note how when you give it only one radius (e.g. `full_coag(radius=[1e-9, 2e-9, 3e-9])`, it returns a 3x3 matrix --- that's because the default is such that, `other_radius=radius`. You can play around with these keyword arguments: `radius` is the only absolutely necessary one; otherwise you will get a customized error like this:


```
>>> full_coag() # or full_coag(temperature=300)
Traceback (most recent call last):
...
    raise ValueError("You must provide a radius!")
ValueError: You must provide a radius!

```

Other keyword arguments:
- essential ones like: `other_radius`, `density`, `other_density`, `charge`, `other_charge`, `temperature`, `pressure`, etc.
- override ones like: `mfp`, `dynamic_viscosity`, `knu`, `sfc`, etc. (these are usually calculated by default, but can be overridden)
- deep-in-the-code ones like: `molecular_weight`, `reference_viscosity`, `reference_temperature`, etc. (these usually replace predefined constants)

ngam:
```python
import numpy as np
from matplotlib import pyplot as plt
from particula import u
from particula.util.dimensionless_coagulation import full_coag

a_rad = np.linspace(1e-9, 1e-6, 1000)

coag_kern = full_coag(radius=a_rad).m_as(u.cm**3/u.s)

plt.loglog(a_rad, coag_kern)
plt.savefig("test.pdf")

```

[test.pdf](https://github.com/uncscode/particula/files/8106024/test.pdf)

ngam:
Do you recognize this shape???

ngam:
Btw, this takes less than a second to run, which is expected. The challenge with coagulation is the time-stepping... which we will be dealing soon enough

ngam:
I am going to hold off doing anything with  classes for now. I need to think about a way to unify them and tightly link them. But you can use the above example to get started.

ngam:
@Gorkowski, now this is clean and completely rebased. I am not sure what purpose is serves. Could you explain to me the goals to help get us there?

Here's my latest thinking on this: 
we need to rethink our object-oriented structure: I am fine with keeping it in general, but we need to start tightly integrating the objects. For this, we need to have a base class somewhere, then build particle from it, and then build particledist on top of that. So these classes will all have to share a common structure and build on top of each other. They will utilize the new flexible methods and so the classes will be housing the new objects simply (but not really any calculations). I need to think harder about how to achieve this --- object-oriented programming is really not my thing, so I don't find it easy to think through this...

Gorkowski:
Here are my goals, to mimic the coagulation in a lofting wildfire plume. Very high particle concentrations near the fire 10s of mg/m3, to dilution in the 3-5 km alt to <100 ug/m3.
 
1) distribution of particles, radii & number
2) run a coagulation step at some specified interval, and update the distribution
3) save out the distribution to plot the mass, and number evolution.
4) parcel mode: dilute the distribution by a constant rate (to mimic mixing with clean air), which will decrease the coagulation rates. Adiabatic parcel path of decreased temp vs pressure at some constant updraft velocity (or prescribed velocity)

ngam:
> 1. distribution of particles, radii & number

So this class will handle this part, right? Now I understand your long-term plan!

Gorkowski:
right, but if we want to refractor with a base class that scales from one to many particles that should be easy to substitute in and replace this particle_distribution class.

ngam:
```python
""" some "new" class
"""

from particula.util.dimensionless_coagulation import full_coag
from particula.util.particle_mass import mass


class ParticleDistribution:
    """ blah
    """

    def __init__(self, **kwargs):
        """ get stuff
        """

        self.kwargs = kwargs

    def mass(self):
        """ get number
        """
        return mass(**self.kwargs)

    def coag(self):
        """ get coag
        """

        return full_coag(**self.kwargs)

```

PR #136, opened 2022-02-22
Issue135

### LINKED ISSUE #135: Time-step of a particle distribution with coagulation
Adding coagulation with time steps.

Given a log-normal particle distribution, 10^9 particles/cc, mode 100 nm, geometric sigma 1.5.

1) Calculate the production and loss of particles at each radius for a given delta time
2) update the distribution with the given production/loss
3) repeat until the specified time ends.

Checks-
1. the total mass concentration should be conserved. 
2. smaller particles are lost to larger ones.

ngam:
Fixes #135


ngam:
@Gorkowski, this is more or less it for #135. I just need to fix this part

https://github.com/uncscode/particula/blob/c03cea704a9d92d17dc3677081b7fa2f062852fb/particula/aerosol_dynamics/dynamic_step.py#L28-L36

by doing a legit coagulation rate calculation (currently, it is fake and wrong)

ngam:
Also this is the correct way to implement the inheritance lol 😆 

https://github.com/uncscode/particula/blob/c03cea704a9d92d17dc3677081b7fa2f062852fb/particula/aerosol_dynamics/dynamic_step.py#L9-L36

ngam:
Wasted a lot of time investigating this part, but the idea is the following:

https://github.com/uncscode/particula/blob/c03cea704a9d92d17dc3677081b7fa2f062852fb/particula/aerosol_dynamics/tests/particle_distribution_test.py#L46-L70

We can trim the distribution after each time step, to keep it small and manageable in the future. The trimming happens via this utility:

https://github.com/uncscode/particula/blob/c03cea704a9d92d17dc3677081b7fa2f062852fb/particula/util/radius_cutoff.py#L8-L18

where we can give it like .9999 (meaning covering all possible range with 99.99% confidence roughly speaking)

ngam:
And this allows us to track these distribution properties in time, so calculating how the mean, geometric std, etc. vary over a process. Very cool!

ngam:
@Gorkowski, i believe this is a decent solution for now. I will want to:
1. clean up the code
2. add more testing
3. explain areas to improve (getting rid of the for loop for example + adding more rigorous stuff for the trimming)
4. documenting the methodology clearly
5. explain how an alternative methodology can be implemented using discrete math (this is the continuous math solution to the problem)

could you have a look when you have a chance and let me know if I am missing anything?

ngam:
This remains a rough implementation. I will need to document stuff, and there is a high chance I will reorganize stuff once I start documenting. I just wanted to get something to work first, then I will start thinking of ways to optimize it and organize it in a meaningful way (e.g. now we are shuttling back and forth between aerosol_dynamics and util, this shouldn't happen imo --- it may introduce circular imports later...)

ngam:
Current state:
- comment dynamic step file
- correct units in dynamic step file (rethink original distribution units?)

ngam:
@Gorkowski depending on where you're at, this can go in. I will follow up on this with issues outlined in #138. I couldn't find a decent solution to bug yet --- it's likely okay for the time being. 

The only solution I found resulted in like 1000x slowdown (see code below) in that for-loop because of the need to go outside of it and call other functions to remedy the situation. So instead of that, I am thinking of either customizing some utilities for this specific problem (expanding matrices, etc.) or I will need to think really hard about a nicer solution.

So anyway, if you feel like this could help you move forward, then merge away (the bug will be fixed later; it is okay for preliminary calculations --- again, slightly less accurate). I will need to review my previous implementation of this to see how I went about setting these for loops and other tricks. I should have another PR for #138 by early next week...


```python
    def coag_gain(self):
        """ Returns the coagulation gain rate

            Equation:

            gain_rate(other_radius) = (
                other_radius**2 *
                integral( # from some_radius=0 to other_radius/2**(1/3)
                    kernel(some_radius, (other_radius**3-some_radius**3)*(1/3) *
                    dist(some_radius) *
                    dist((other_radius**3 - some_radius**3)*(1/3)) /
                    (other_radius**3 - some_radius**3)*(2/3),
                    some_radius
                )
            )

            Units:
                m**-4/s

            This equation necessitates the use of a for-loop due to the
            convoluted use of different radii at different stages.
            This is the costliest step of all coagulation calculations.
            Note, to estimate the kernel and distribution at
            (other_radius**3 - some_radius**3)*(1/3)
            we use interporlation techniques.

            This could be made better in the future.
        """

        # get the parameters
        nums, rads, _ = self.coag_prep()

        # set a radius value
        dps = rads

        # make a sekeloton array for the gain
        gain = np.zeros_like(dps)

        # predefine function coag:
        get_coag_kern = full_coag
        # loop over the radius of interest (dps above)
        for i, dpa in enumerate(dps):

            # make the dummy radius for integration
            dpd = np.linspace(0, dpa/2**(1/3), 1000)

            # get the dummy radius for interpolation
            dpi = (dpa**3 - dpd**3)**(1/3)

            # interpolate the distribution to the dummy radii
            num_rep = np.interp(dpd, dps, nums)
            num_oth = np.interp(dpi, dps, nums)

            # interpolate the kernel to the dummy radius
            # call the kernel function again; HUGE SLOWDOWN
            ker_oth = get_coag_kern(
                radius=dpd, other_radius=dpi, **self.kwargs
            )

            # only need diag inside the loop
            ker_oth = np.diag(ker_oth.m)*ker_oth.u

            # get rid of nan due to radius=0 in dpi, dpd
            ker_oth = np.nan_to_num(ker_oth)

            # calculate last term
            dss = (dpa**3 - dpd**3)**(2/3)

            # calculate the gain
            test = (dpa**2)*np.trapz(ker_oth*num_oth*num_rep/dss, dpd)

            # store the gain
            gain[i] = test.m

        # return it
        return gain*test.u

```

Gorkowski:
So I'm testing how this would work in practice in a jupyter notebook, on colab.
https://colab.research.google.com/drive/1BGbUAAT_A9iR2_dmItqTugM6AI7LBeT_?usp=sharing

The main question's in there are
1) why are the particle distribution units 1/m^4
2) why is the coagulation loss so high, as after the first timestep you have negative values? or maybe my understanding of how to use this is wrong.

ngam:
> So I'm testing how this would work in practice in a jupyter notebook, on colab. https://colab.research.google.com/drive/1BGbUAAT_A9iR2_dmItqTugM6AI7LBeT_?usp=sharing
> 

You need to make this public.

> The main question's in there are
> 
> 1. why are the particle distribution units 1/m^4

Because it is a concentration density, not pure concentration. Very briefly if you go back to the textbooks on this you will see a more rigorous explanation, but for me personally, it is just a visual trick --- that extra m you see is just the radius, so you can multiply/divide by it to see how things will look differently. The important thing is to keep track of this when ones does integration. Think: Integrate( distribution over radius ) == ( distribution * radius )  over some interval, so you see how you need the extra radius unit in there to return back a meaningful unit for the distribution. This goes back to the confusion I mentioned about legit probability density functions (key: "density") vs probability functions.

> 2. why is the coagulation loss so high, as after the first timestep you have negative values? or maybe my understanding of how to use this is wrong.

This is suspicious --- it could be something wrong in the calculation, but generally it should work relatively well. CoagLoss rate = d Distribtuon / d t = - Distribution * integral (Distribution * Kernel,  over radius), this will be asymptotically zero / stable. Meaning: d Distribution / d t can be rewritten as  ( 1 / Distribution ) ( d Distribution / dt ) = something negative  which is always stable ---> Distribution will have an exponential function of the form exp (- something negative * t). So your intuition is, of course, correct. It is either there is something wrong in the code or your way of doing it...

ngam:
I think this is wrong:

https://github.com/ngam/particula/blob/54bfb86ce93edfe376f62c3ccd6f4fb55e770b31/particula/aerosol_dynamics/particle_distribution.py#L103

I should fix this later today.

ngam:
@Gorkowski, I fixed the issue. Note, that now you will get very tiny coagulation (es expected!) so you should increase the time step. Also now, we deal with density but project the unit as 1/L3 (I will need to clear this up later, but for now, it should be okay --- I will want to ensure the math is very solid before we move forward)

The issue was that my way of "clarifying" the whole mess I describe below was actually somehow really messing things up. I thought I could be smarter than it by passing (1/radius) somewhere but for some bizarre reason that's not readily clear to me, that actually completely corrupts all calculations. I was already kind of confused about why the masses weren't conserved as nicely as I wanted them to be,  so previously I had `pytest.approx(0.0, abs=1e-8)` and I was like I will look into it later as it should be really really close to 0 in my calculation. Turns out I should've paid much more attention to this subtle hint. in the updated code, I have this as `pytest.approx(0.0)` and it passes!

> > So I'm testing how this would work in practice in a jupyter notebook, on colab. https://colab.research.google.com/drive/1BGbUAAT_A9iR2_dmItqTugM6AI7LBeT_?usp=sharing
> 
> You need to make this public.
> 
> > The main question's in there are
> > 
> > 1. why are the particle distribution units 1/m^4
> 
> Because it is a concentration density, not pure concentration. Very briefly if you go back to the textbooks on this you will see a more rigorous explanation, but for me personally, it is just a visual trick --- that extra m you see is just the radius, so you can multiply/divide by it to see how things will look differently. The important thing is to keep track of this when ones does integration. Think: Integrate( distribution over radius ) == ( distribution * radius ) over some interval, so you see how you need the extra radius unit in there to return back a meaningful unit for the distribution. This goes back to the confusion I mentioned about legit probability density functions (key: "density") vs probability functions.
> 
> > 2. why is the coagulation loss so high, as after the first timestep you have negative values? or maybe my understanding of how to use this is wrong.
> 
> This is suspicious --- it could be something wrong in the calculation, but generally it should work relatively well. CoagLoss rate = d Distribtuon / d t = - Distribution * integral (Distribution * Kernel, over radius), this will be asymptotically zero / stable. Meaning: d Distribution / d t can be rewritten as ( 1 / Distribution ) ( d Distribution / dt ) = something negative which is always stable ---> Distribution will have an exponential function of the form exp (- something negative * t). So your intuition is, of course, correct. It is either there is something wrong in the code or your way of doing it...

ngam:
>  I thought I could be smarter than it by passing (1/radius) somewhere but for some bizarre reason that's not readily clear to me, that actually completely corrupts all calculations.

Ah, scipy.stats.lognorm already divides by radius! That's why... LOL, RIP!

Gorkowski:
Things look good now, I'll merge the PR and work on the notebook covering this initial method

download (1)

Gorkowski (approved):

ngam:
Sounds good, the documentation for now should target @wkchuang and @aktwo only. Still we will likely tweak the implementation once we decide how we should do the object-oriented stuff (#137).

Looking forward to seeing your time-stepping, but remember ... we ought to have a proper solver soon! I will write a quick 2nd order solver as a utility soon-ish!

PR #158, opened 2022-03-04
improve testing for logspace stuff

### LINKED ISSUE #157: improve testing based on recent bugs
Two bugs should've been caught by better testing (#154 and #153).

For #154: we need stricter bounds on the mass conservation, and we should make it relative to gain/loss (rather than absolute) as the absolute can be misleading

For #155: well, we need to actually have tests for this; the issue here is that logspace is really bad for coagulation calculation --- however, we do have guardrails against then in the coagulation calculation itself: in calculating gain (where it matters most), we impose using linspace anyway and so the outside (apparent) spacing shouldn't matter and the user should be okay using either one without any cost penalty (for the time being)

ngam:
Fixes #157

### DIFF
--- a/particula/aerosol_dynamics/tests/particle_distribution_test.py
+++ b/particula/aerosol_dynamics/tests/particle_distribution_test.py
@@ -20,17 +20,29 @@ def test_rad():
 
     pdist = ParticleDistribution(
         cutoff=.9999,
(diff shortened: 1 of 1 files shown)


Gorkowski (approved):
I'll always support better tests!!

PR #167, opened 2022-03-08
make environment the base class

### LINKED ISSUE #166: make environment the base class
rewrite env class to make it base and build particle on top of it

ngam:
Fixes #166

Idea:
`Environment` will be our base class, for now taking only* two inputs (pressure, temperature) and outputting four things (pressure, temperature, viscosity, mfp). Then, we build Particle on top of Environment, thus inheriting all the above without having to re-input/re-output them. This will make Environment more or less a hidden class. We can expand the inputs and outputs of Environment as we see fit later, and as we organize the other classes on top of it. 

To prepare you for the way ahead: Particle will inherit from Environment; Then ParticleDistribution will either be blended into Particle or will be built on top of it; then CoagulationRate, CondensationRate, etc. are built on top of ParticleDistribution. However, we will likely need another class for Vapor, but I will need to think more about where in the sequence that belongs. It will likely have to be right before Particle because in reality, particles come from Vapor components (nucleation + condensation), so "etymologically" something like this would make most sense, but Vapor treatment is pretty advanced and will take some time. By vapor(s), I mean gaseous precursors (but not the medium itself; the medium is Environment).

`Environment > Vapor > Particle > Rates > ..` 

Highlights:
- relies on utils instead of in-class methods
- ensures multi-dim inputs work correctly
- removed units checks (relying instead on input_handling utils)
- removed `_air` from names as the utils actually apply to more than air (but for all practical purposes, we will be dealing only with air)

TODO:
- [x] decide if to document the kwargs stuff here (it is fully documented in utils)
- [x] add more rigorous testing
- [x] remove getters (no need to output pressure/temperature anymore; requires fixes in downstream in particle, parcel)

DO NOT MERGE as we should finalize some things before merging.

NOTE:
* *in reality, kwargs allow for many more inputs including overriding utils and constants. See added notes atop the class.


wkchuang:
> Fixes #166
> 
> Idea: `Environment` will be our base class, for now taking only* two inputs (pressure, temperature) and outputting four things (pressure, temperature, viscosity, mfp). Then, we build Particle on top of Environment, thus inheriting all the above without having to re-input/re-output them. This will make Environment more or less a hidden class. We can expand the inputs and outputs of Environment as we see fit later, and as we organize the other classes on top of it.
> 
> To prepare you for the way ahead: Particle will inherit from Environment; Then ParticleDistribution will either be blended into Particle or will be built on top of it; then CoagulationRate, CondensationRate, etc. are built on top of ParticleDistribution. However, we will likely need another class for Vapor, but I will need to think more about where in the sequence that belongs. It will likely have to be right before Particle because in reality, particles come from Vapor components (nucleation + condensation), so "etymologically" something like this would make most sense, but Vapor treatment is pretty advanced and will take some time. By vapor(s), I mean gaseous precursors (but not the medium itself; the medium is Environment).
> 
> `Environment > Vapor > Particle > Rates > ..`
> 
> Highlights:
> 
>     * relies on utils instead of in-class methods
> 
>     * ensures multi-dim inputs work correctly
> 
>     * removed units checks (relying instead on input_handling utils)
> 
>     * removed `_air` from names as the utils actually apply to more than air (but for all practical purposes, we will be dealing only with air)
> 
> 
> TODO:
> 
>     * [ ]  decide if to document the kwargs stuff here (it is fully documented in utils)
> 
>     * [ ]  add more rigorous testing
> 
>     * [ ]  remove getters (no need to output pressure/temperature anymore; requires fixes in downstream in particle, parcel)
> 
> 
> DO NOT MERGE as we should finalize some things before merging.
> 
> NOTE:
> 
>     * *in reality, kwargs allow for many more inputs including overriding utils and constants. See added notes atop the class.

Sounds good! I like how we're organizing the hierarchy. I'll take a closer review tonight.

ngam:
Thanks @wkchuang, please think about where and how we should organize the inputs/outputs. 

I am leaning towards the idea to never let utilities access **kwargs and instead they should be defined explicitly. For example,

dynamic_viscosity.py currently has:
```python
def dyn_vis(**kwargs):

    temp = kwargs.get("temperature", 298.15)
    ref_vis = kwargs.get("reference_viscosity", REF_VISCOSITY_AIR_STP)
    ref_temp = kwargs.get("reference_temperature", REF_TEMPERATURE_STP)

    temp = in_temperature(temp)
    ref_vis = in_viscosity(ref_vis)
    ref_temp = in_temperature(ref_temp)

    suth_const = SUTHERLAND_CONSTANT

    return (
        ref_vis * (temp/ref_temp)**(3/2) * (ref_temp + suth_const) /
        (temp + suth_const)
    )
```

instead, we move the kwargs from that file to the environment.py file, and we define stuff explicitly inside the utility:

dynamic_viscosity.py will have:

```python
def dyn_vis(temp=298.15*u.K, ref_vis=..., ref_temp=..., suth_const=...,):

    return (
        ref_vis * (temp/ref_temp)**(3/2) * (ref_temp + suth_const) /
        (temp + suth_const)
    )
```

ngam:
This cosmetic may be beneficial when it comes to never having to check inside the utilities again (there's a reason they're hidden in util, right?) but I am not 100% sure. So I would like to hear from you what makes more sense in terms of organizing inputs/outputs. Main idea: never repeat an input.

wkchuang:
> Thanks @wkchuang, please think about where and how we should organize the inputs/outputs.
> 
> I am leaning towards the idea to never let utilities access **kwargs and instead they should be defined explicitly. For example,
> 
> dynamic_viscosity.py currently has:
> 
> ```python
> def dyn_vis(**kwargs):
> 
>     temp = kwargs.get("temperature", 298.15)
>     ref_vis = kwargs.get("reference_viscosity", REF_VISCOSITY_AIR_STP)
>     ref_temp = kwargs.get("reference_temperature", REF_TEMPERATURE_STP)
> 
>     temp = in_temperature(temp)
>     ref_vis = in_viscosity(ref_vis)
>     ref_temp = in_temperature(ref_temp)
> 
>     suth_const = SUTHERLAND_CONSTANT
> 
>     return (
>         ref_vis * (temp/ref_temp)**(3/2) * (ref_temp + suth_const) /
>         (temp + suth_const)
>     )
> ```
> 
> instead, we move the kwargs from that file to the environment.py file, and we define stuff explicitly inside the utility:
> 
> dynamic_viscosity.py will have:
> 
> ```python
> def dyn_vis(temp=298.15*u.K, ref_vis=..., ref_temp=..., suth_const=...,):
> 
>     return (
>         ref_vis * (temp/ref_temp)**(3/2) * (ref_temp + suth_const) /
>         (temp + suth_const)
>     )
> ```
> 
> update: see example implementation in [c94023c](https://github.com/uncscode/particula/pull/167/commits/c94023cd66932e42c6d23d87d41f44fb94da21d4)

I like it this way, where the important inputs are explicitly stated so it's clearer to the reader.

PR #176, opened 2022-03-13
wip: reformulate particle class to follow env > vap ..

### LINKED ISSUE #175: reformulate particle class to follow environment > vapor ...

ngam:
Fixes #175


ngam:
@wkchuang + @Gorkowski do you have a good feeling for how the class stacking works now?

For the particle class, I am splitting it into two (potentially more) to better organize things. For now, BaseParticle will build on top of Vapor, and will only take basic particle properties. Particle builds on BaseParticle to output particle--particle stuff (reduced mass, coagulation, etc.). Once I have clarity about building condensation in a reasonable way, it will be added to Particle. I will need to trim down and organize the utils in a more coherent way so that we don't have circular imports, dependencies. Hence, still WIP (work in progress) / Draft. But I think this is largely done for now (I will need to integrate better testing and better "distribution" input).

While stacking is important, I don't want us to be hamstrung to this setup in particular so I want to limit the output of these classes only basic derived properties (e.g. coagulation) because I intend to have a full-fledged coagulation class based on Particle class soon that will have thorough (and independent; to avoid circularity) details with more fine-grained and step-wise details. Anyway, an important point in constructing these classes will be flexibility, so I expect we will go back and revise them once we gather more details about how downstream steps look like...

particula/util/particle_mass.py
```diff
@@ -33,18 +39,13 @@ def mass(**kwargs):
         Returns:
                          (float) [kg]
     """
-
-    radius = kwargs.get("radius", None)
-    density = kwargs.get("density", 1000)
-    shape_factor = kwargs.get("shape_factor", 1)
-    vol_void = kwargs.get("volume_void", 0)
-
+    _ = kwargs.get("something", None)
```
Gorkowski, line 42: what is this? "something"?

Gorkowski (changes requested):

particula/util/particle_mass.py
ngam: LOL, this was just to avoid upsetting the linter. The idea is that we should pass `kwargs` through, but we don't really need it here. If you don't use it, it will throw a linter error. This trick `_ = kwargs.get("something", None)` will use `kwargs` to read a keyword argument called "something", defaulting to `None`. Note that, effectively, `_` means it is just thrown out regardless and not used regardless; "something" here could be really anything. I can also just get rid of this and disable the linter warning/error instead.

Gorkowski: ha, I support the passing of kwargs everywhere :) as we build things out. so we're good here

Gorkowski (approved):

PR #183, opened 2022-03-13
revise environment class to return (1, 1, x)

### LINKED ISSUE #172: revise environment class to return (1, 1, x)
That is, make the temperature dimension the third dimension. So if `{..., "temperature": [300, 301], ...}`, it should be returned as (e.g. via `self.temperature`) as a ndarray of shape (1, 1, 2).

ngam:
Fixes #172


ngam:
I don't think this is a good idea for now, so closing.

PR #191, opened 2022-03-26
integrate particle_distribution into particle

### LINKED ISSUE #190: integrate particle_distribution into particle

### LINKED ISSUE #196: cleanup particle class
1. allow for particle--particle default resolution:

    - This can be achieved by changing setting a default for other, `other: "Particle" = None` and then calling with `(other or self)`.

3. rename BasePreParticle and BaseParticle to what they actually mean: 
   - BasePreParticle = instantiate Particle from a continuous pdf distribution 

   - BaseParticle = instantiate Particle from given radii and variables AND calculates particle (singular) properties (maybe should split this?)

   - Particle = calculates particle--particle (plural) properties

ngam:
Fixes #190
Fixes #196


Gorkowski (approved):
good upgrade!!

PR #193, opened 2022-03-27
getting rid of particula.units

### LINKED ISSUE #173: particula.units is obsolete
With the introduction of `particula.util.input_handling`, we ensure that we are always dealing with quantities with units (i.e. `particula.u.Quantity`). The strip utility in `particula.units` is thus obsolete as we can always take `.m` of any `particula.u.Quantity` with no issues.

ngam:
Fixes #173


Gorkowski (approved):
winning!

PR #216, opened 2022-04-13
Give some examples of coagulation

### LINKED ISSUE #215: Give some examples of coagulation
The code is pretty dense, can we get an example of how to coagulate particles together. 

Could this be in a notebook :)

Gorkowski:
Fixes #215

adds coagulation notebook


ngam:
I think you ran this locally and ended up adding the main package as a submodule.

Also, 

1. how do you feel about putting this in a folder called "tutorials" / "examples" / "demos" ?
2. you could build the docs locally* and see how the html looks like if you want, or we can just merge this and see it live --- let me know which you prefer

*building locally:
```
$ pip install ".[dev]"
$ jupyter-book build docs

```

ngam (approved):
The only minor suggestion is to move coagulation to a new folder called demos, but we can do that later. I will approve this for now. Leave it up for a day or two in case you want to add anything, then merge. (Note that there are few minor typos and clarification, but I prefer we do these in following PRs instead of now --- it will also give you a chance to see the added documentation live)

ngam:
Another thing, I usually do a lot of "pre_" stuff, like "pre_radius" and "pre_distribution", but the user is not supposed to actually call these. Unless otherwise given elsewhere, radius and distribution will fall back on pre_radius and pre_distribution. It's just a trick to set defaults. Maybe I/we should consider making the names more obvious or eliminating these intermediates altogether?

Gorkowski:
I'm going to look at this again Sunday so, hold off the merge 😁

Gorkowski:
@ngam OK, I reomved the cloning junk at the top and added this under examples.

PR #217, opened 2022-04-19
allow/combine multiple-species condensation

### LINKED ISSUE #206: allow multiple species condensation
For example, allow two condensing vapor. I believe it's best to calculate an effective gross condensation rate (in the final calculation) rather than treat them separately. This will make the solver happier. Should document implementation clearly.

ngam:
fixes #206 

This PR reorganized the condensation calculation so that it is possible to sum all condensation contributions into one driving force (i.e. diameter growth). This involves refactoring one function into a utility and transposing matrices to make it work.


particula/util/vapor_flux.py
```diff
@@ -0,0 +1,52 @@
...
+                                           in_scalar)
+from particula.util.molecular_enhancement import mol_enh
+from particula.util.particle_surface import area
+from particula.util.rms_speed import cbar
+
+
+def phi(  # pylint: disable=too-many-arguments
+    particle_area=None,
+    molecular_enhancement=None,
+    vapor_attachment=1,
```
Gorkowski, line 17: is vapor attachment, the same as the accommodation coefficent?

ngam: In short, yes, but we should generally distinguish between different attachment coefficients. There are gonna be particle--particle attachment coefficients, and this is the vapor--particle coefficient. Once we start doing "chemistry" later on, these may need some more thought...

Gorkowski (approved):
nice!

PR #312, opened 2022-10-17
add kwargs to solver

ngam:
@Gorkowski use do_coag, do_cond, do_nucl for now. We can fix this and do it more properly later. Releasing soon.


Gorkowski:
Yeah that's a good plan, I'm going to think of some more changes that would be good after work with this for class

PR #313, opened 2022-10-22
add condensation equations

### LINKED ISSUE #307: resolve condensation
Let's go deeper into our treatment of condensation growth to resolve all the terms carefully. Copying @Gorkowski

> From this https://www.nature.com/articles/nature18271, go to "aerosol growth model" --- I envision dynamically resolving that can resolve things. I will ask NMD about it out of curiosity, "can this be applied across scales, e.g. for droplet activation?"

> _Originally posted by @ngam in https://github.com/uncscode/parcelona/issues/9#issuecomment-1273433210_

ngam:
fix #307 

@Gorkowski can you help me get values for these parameters? What is usually known and what is not known? Where to look?

| parameter | units | status |
| --------- | ----- | ------ |
| vapor_concentration | #/cc | ok |
| sat_vapor_concentration | #/cc? | ? |
| mass_fraction (X) | unitless? | ? |
| activity_coefficient | unitless? | ? |

@wkchuang this is from your paper, not sure if you recall any of this stuff or if you were the one working on it back in the day...


Gorkowski:
> We calculate the vapor concentration from vapor pressure. Saturation vapor pressure is parameterized in terms of temperature, and the vapor pressure can then be assumed to drive condensation.

We can decide if we want to Ci, C* tracking instead of pi, p*. If we can add custom units, I think C* is more intuitive (if your a Neil student :) than p*, but p* is better when looking at literature databases for specific compounds. Either way, we can add a function to convert between the with mass.

> We can calculate mass fraction based on how much water made it into the particle, but I am not sure this is the way to go. Water activity can be broken down into elemental parts based on KG's McG paper.

Yes, I think this is the way to go. This way we get the current activity and can compare it to the gas phase. 

> Now my remaining questions: how many of these parameters must be dynamic?
> * mass fraction must be recalculated after time steps (changes as condensation goes on)
> * we can assume the temperature constant, but that's actually likely wrong, there will be some thermal transfer due to this (easy to formulate); this will mean sat vapor pressure is also changing
> * vapor pressure will also decrease as condensation goes (depletion)

activity coefficient for water is dynamic (temperature (weak) and composition (strong))
pure vapor pressure, should be a cheap calculation, but it is dynamic.

> is it safe to assume the activity is constant under all conditions or are there interesting parameterizations here?

I think for now, we can use a look up table for water activity coefficients with specific organic compounds (NaCl, Sulfate,...). Mass or mole fraction to activity coefficient.

> In other words, the fundamental independent variables are:
> 
> * vapor pressure (thus, dependent variable is mass fraction), so need: dT/dt
> * temperature (thus, dependent variable is sat vapor pressure), so need: dP/dt
> * activity

* activity = activity coefficient * mass fraction
* and activity coefficient is a function of mass fraction and temperature

Gorkowski:
I'm pulling in some code from McGill here for the CCN formation. My plan is to pull in the kappa framework, and then integrate it to use our particle class, for condensation rate calculations of water. 

I'm still pondering the best way to add a wet radius or just water mass/volume.

Gorkowski:
Next up is accounting for more than one material in the particle/vapor phase

PR #325, opened 2023-01-11
expose modules

### LINKED ISSUE #324: better modules
properly expose main modules, so instead of `particula.particle.Particle`, we should have `particula.Particle`

ngam:
fix #324


Gorkowski (approved):
thats a bunch of small changes :), but I do like it better this way

PR #351, opened 2023-11-12
Gorkowski/issue343

Gorkowski:
starting with just the BAT activity coefficient calculation for now


Gorkowski:
this is kinda a big merge, I'll double check things in the morning before sending it.

Gorkowski:
doubled-checked that the coagulation simulation still works, and we didn't break it.

PR #404, opened 2023-11-21
Import example chamber data

### LINKED ISSUE #403: Import example chamber data

Gorkowski:
Fixes #403

- [x] add data filtering
- [x] relative time calc


Gorkowski:
@ngmahfouz part 1 of chamber analysis done.
Load and clean the data notebook.

Gorkowski:
> Interesting this doesn't happen automatically... I thought it would  

me too. I just asked gpt, and numpy only will auto add the dim to the left. "For example, if you have a 2D array A with shape (3,4) and you want to multiply it by a 1D array B with shape (4,), NumPy will automatically broadcast B to have a shape of (1,4), and then stretch it to match the shape of A for the multiplication."

If A has a shape of (3, 50) and B has a shape of (3,), NumPy broadcasting will not be able to automatically align and broadcast these arrays for element-wise operations like multiplication, because their dimensions do not align properly according to the broadcasting rules.

In broadcasting, the trailing dimensions of the arrays are compared, and one of the dimensions has to be 1 for broadcasting to occur. In this case:

A has a shape of (3, 50).
B has a shape of (3,).
When attempting to align these shapes, we get:

(3, 50)
(3, )
The last dimensions (50 from A and the implicit 1 from B) are not compatible because neither is 1 nor are they equal. Therefore, a broadcasting error will occur if you try to directly multiply A and B.

To perform element-wise multiplication between A and B, you would need to reshape B so that it has a compatible shape. For instance, if you want to multiply each row of A by B, you can reshape B to have a shape of (3, 1).

PR #406, opened 2023-11-24
pyTorch explore for Lagrangian

Gorkowski:
Initial exploration into the Lagrangian model, with a target to run on GPUs  (so pytorch) with high particle numbers. Using torch still allows development on CPUs, using identical code.

- [x] Direct coagulation calculation
- [x] simple leapfrog velocity/position integration
- [x] Simple/unrealistic example to show code runs
- [x] Tests for basic Lagrangian utility operations/functions


Gorkowski (commented):
@ngmahfouz What do you think about waiting a bit, for 3.12? It has only been for like two months....so pytorch and pytz haven't supported it yet.

Gorkowski:
> > @ngmahfouz What do you think about waiting a bit, for 3.12? It has only been for like two months....so pytorch and pytz haven't supported it yet.
> 
> We can wait, no problem at all. How did we not catch the pytz error btw?
> 
> 

It's just a warning right now, I think. From issue

Gorkowski:
> Worth mentioning that JAX can be a drop-in replacement for numpy if needed btw, likely with less code changes than pytorch. We can also potentially support different backends...
> 
> see an example of switching back and forth between numpy, numba, and jax: https://github.com/uncscode/hypersolver/blob/main/hypersolver/util.py (both numbe and jax are inactive now)
@ngmahfouz 
I did some research on JAX vs PyTorch. The one sticking point I have right now with JAX is the lack of GPU on Windows support, which Pytorch does have. I don't want to add additional hurdles to new Ph.D students (and myself) if there isn't a need. They both have a compile feature.

For Pytorch on Mac OS, it looks like the default install supports MPS on Apple Metal. I am not sure how that affects things on the code side, I assume just a device selection keyword like CUDA vs CPU vs MPS.

For us, I don't see a major problem if we write it in pytorch, then in v2 write it in JAX/numpy. I would be ecstatic if we had enough users that a v2 was called for :)

PR #424, opened 2024-01-02
Wall loss example

### LINKED ISSUE #402: Chamber Wall Loss Scoping Study
Development of the particle wall-loss terms for a steel smog chamber. 

Stainless Steel Box
Length:  739.14 mm
Width:   607.64 mm
Height:  1663.7 mm

Gorkowski:
Fixes #402

- [x] adding rectangular beta0
- [x] testing rectangular code
- [x] example of different losses


particula/util/wall_loss.py
```diff
@@ -9,41 +9,147 @@
...
+    ktp_value,
+    diffusion_coefficient_value,
+    chamber_radius,
+    settling_velocity_value
+):
+    """Calculate the wall loss coefficient for a spherical chamber
+        approximation.
+
+    Args:
+        ktp_value: ?? turbulent parameter ??
```
Gorkowski, line 22: @ngmahfouz is this the same kt as you wrote in the issue, and what does it mean? or is it just some aerosol specific parameter.

Gorkowski:
I wrote things out more explicitly in wlc, in this update too

Gorkowski (commented):
@ngmahfouz  Yeah, feel free to refactor and add to this PR. I was thinking of protyping out the wall loss forward simulation. 

After that work on the data fitting as a seperate PR.

Gorkowski:
@ngmahfouz Here is my first cut of the chamber forward simulations. Look at the plots in the notebook to double-check the dynamics look right. 

Also, some negative concentration values slip in at high concentrations of 1e5 and smaller particle sizes. This is due to the ODE taking too large of a step. Ideas on fixing that would be good :)  It's not a significant problem, but something that a novice person could find puzzling if they don't understand it.

PR #427, opened 2024-01-08
Gorkowski/issue426

Gorkowski:
Adding Wall loss fitting via brute force, on a single experiment example.

- [x] Manual comparison
- [x] Auto minimization
- [x] MAE fit


Gorkowski:
@ngmahfouz 

I decided to just show a single time window fit for now, so that my student can start using this. If the functions in the notebook get more used then maybe we pull it in to the code base, but I'm fine with leave them in the example for now.

PR #453, opened 2024-05-11
Pre-Condensation PR gas properties

Gorkowski:
**Summary of Changes in this PR:**

- **Refactored Gas Properties:**
  - Moved gas properties functions into a dedicated `gas.properties` directory to improve locational reference and accessibility.
  - This restructuring aims to reduce code bloat in the `utils` module of version 1 by compartmentalizing functions related to gas properties.

- **Updated Logger Configuration:**
  - Modified the logging setup to exclusively capture logs from `particula`. 
  - This change addresses the issue of extraneous debug logs from `matplotlib` cluttering the log files, ensuring cleaner and more relevant log outputs.

- **Enhanced Reduced Quantity Calculations:**
  - Introduced two new functions in the `reduced_quantity` module to handle calculations for both individual and array-based inputs:
    - One function is tailored for array to array calculations.
    - The second function is designed for self operations of arrays, enhancing computational efficiency and utility for batch processing.

### DIFF
--- a/particula/next/gas/properties/dynamic_viscosity.py
+++ b/particula/next/gas/properties/dynamic_viscosity.py
@@ -1,0 +1,59 @@
+""" Module for calculating the dynamic viscosity
+
+    The dynamic viscosity is calculated using the Sutherland formula,
+    assuming ideal gas behavior, as a function of temperature.
+
+    "The dynamic viscosity equals the product of the sum of
+    Sutherland's constant and the reference temperature divided by
+    the sum of Sutherland's constant and the temperature,
+    the reference viscosity and the ratio to the 3/2 power
+    of the temperature to reference temperature."
+
+    https://resources.wolframcloud.com/FormulaRepository/resources/Sutherlands-Formula
+"""
+
+import logging
+from particula.constants import (REF_TEMPERATURE_STP, REF_VISCOSITY_AIR_STP,
+                                 SUTHERLAND_CONSTANT)
+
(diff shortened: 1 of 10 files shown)


particula/next/gas/properties/dynamic_viscosity.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Raises:
+    ------
+    - ValueError: If the temperature is less than or equal to 0.
+
+    References:
+    ----------
+    https://resources.wolframcloud.com/FormulaRepository/resources/Sutherlands-Formula
+    """
+    if temperature <= 0:
```
Gorkowski, line 52: I thought about abstracting this to a function, but then the logger would save the location of the function error, and not log the error as from get_dyanmic_viscosity LineXXX.

I'm thinking of keeping this this way, until we figure out the abstraction correctly. Which may mean passing more arugments or something.

particula/next/gas/properties/dynamic_viscosity.py
```diff
@@ -0,0 +1,59 @@
...
+        logger.error("Temperature must be greater than 0 Kelvin.")
+        raise ValueError("Temperature must be greater than 0 Kelvin.")
```
Gorkowski, line 54: I thought about abstracting this to a function, but then the logger would save the location of the function error, and not log the  error as from get_dyanmic_viscosity LineXXX.

I'm thinking of keeping this this way, until we figure out the abstraction correctly. Which may mean passing more arguments or something.

PR #456, opened 2024-05-16
add github citation button

Gorkowski:
Adding the citation button, on the GitHub repo home page.

see, 
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files


Gorkowski:
> I don't think we should add this before v0.1.0 ... Let's stabilize first

That's a good idea 👍

PR #461, opened 2024-05-29
Pre-Condensation 0 Builder ABC class

Gorkowski:
**Changes Implemented:**
- **BuilderABC Class:**
  The `BuilderABC` class abstracts common methods for building classes into an Abstract Base Class, such as `check_keys`, `set_parameters`, `pre_build_check`, and `build`. This ensures consistency and reusability across different builder classes. For example, the vapor pressure builder and factory have been updated to utilize `BuilderABC`, removing code duplication.

- **Input Handling (`convert_units`):**
  The `convert_units` function in input handling has been improved to manage non-multiplicative unit conversions. An additional parameter `value` was added to support conversions, such as temperature, which are not simple scaling by multiplication.

Example usage, from the `Future`, for how mixin's are used to add methods dynamically to builders.

```python
class SurfaceStrategyMolarBuilder(
    BuilderABC,
    BuilderDensityMixin,
    BuilderSurfaceTensionMixin,
    BuilderMolarMassMixin
):
    """Builder class for SurfaceStrategyMolar objects.

    Methods:
    --------
    - set_surface_tension(surface_tension, surface_tension_units): Set the
        surface tension of the particle in N/m. Default units are 'N/m'.
    - set_density(density, density_units): Set the density of the particle in
        kg/m^3. Default units are 'kg/m^3'.
    - set_molar_mass(molar_mass, molar_mass_units): Set the molar mass of the
        particle in kg/mol. Default units are 'kg/mol'.
    - set_parameters(params): Set the parameters of the SurfaceStrategyMolar
        object from a dictionary including optional units.
    - build(): Validate and return the SurfaceStrategyMolar object.
    """

    def __init__(self):
        required_parameters = ['surface_tension', 'density', 'molar_mass']
        BuilderABC.__init__(self, required_parameters)
        BuilderSurfaceTensionMixin.__init__(self)
        BuilderDensityMixin.__init__(self)
        BuilderMolarMassMixin.__init__(self)
```

### DIFF
--- a/particula/next/abc_builder.py
+++ b/particula/next/abc_builder.py
@@ -1,0 +1,287 @@
+"""Abstract Base Class for Builder classes.
+
+This module also defines mixin classes for the Builder classes to set
+some optional method to be used in the Builder classes.
+https://en.wikipedia.org/wiki/Mixin
+"""
+# pylint: disable=too-few-public-methods
+
+from abc import ABC, abstractmethod
+from typing import Any, Optional, Union
+import logging
+from numpy.typing import NDArray
+import numpy as np
+
+from particula.util.input_handling import convert_units  # type: ignore
+
+logger = logging.getLogger("particula")
+
+
+class BuilderABC(ABC):
+    """Abstract base class for builders with common methods to check keys and
+    set parameters from a dictionary.
+
+    Attributes:
+    ----------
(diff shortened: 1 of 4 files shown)


Gorkowski:
@ngmahfouz  <500

PR #498, opened 2024-10-09
Test for py13

mahf708:
## Summary by Sourcery

Update CI workflows to use Python 3.12, add Python 3.13 to the test matrix, and adjust the 'pint' dependency version.

Enhancements:
- Update the dependency version for 'pint' to be at least 0.24.4 in the project configuration.

CI:
- Update CI workflows to use Python 3.12 instead of Python 3.11 for linting and documentation generation.

Tests:
- Add Python 3.13 to the testing matrix in the CI workflow.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,10 +20,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     
-    - name: Set up Python 3.11
+    - name: Set up Python 3.12
       uses: actions/setup-python@v5.3.0
       with:
-        python-version: "3.11"
+        python-version: "3.12"
         cache: 'pip' # caching pip dependencies
         
     - name: Install dependencies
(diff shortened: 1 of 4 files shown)


Gorkowski:
@mahf708 It is pint, that is the problem and not yet supporting 3.13

https://github.com/hgrecco/pint/issues/1969

Gorkowski:
> Do we still plan to keep pint for the future of particula? I thought there was enough momentum to get rid of it ...

Um, not really. The units inside the calculations are now all base SI. But to setup the model I still allowed for setting units when building the different objects. They are mostly isolated in the BuilderMixins https://github.com/uncscode/particula/blob/main/particula/builder_mixin.py#L16

and all use the convert_units function https://github.com/uncscode/particula/blob/main/particula/util/input_handling.py
So we could replace convert_units if we want...or isolate it in a better way.

Gorkowski:
@mahf708 we could use https://docs.scipy.org/doc/scipy/reference/constants.html#length as starting point and add extra mappings

Gorkowski (approved):
nice that they updated pint

PR #500, opened 2024-10-14
get rid of numerical issue

mahf708:
dividing by something small can lead to all sorts of issues; when COULOMB_POTENTIAL_RATIO_ARRAY is 0, the logic in next picks 1e-16 as replacement, but that doesn't work for all packaged numpy pkgs (e.g., it was producing nans on conda-forge). This is temporary fix; in the future, we should likely avoid hardwiring stuff like this and think of more creative unit tests... (e.g., calculating stuff from different angles, etc.)

also raising version to 0.0.20 to prep for yet another release

## Summary by Sourcery

Fix division by zero issue in COULOMB_POTENTIAL_RATIO_ARRAY by removing zero value and update test cases accordingly. Increment project version to 0.0.20.

Bug Fixes:
- Remove zero value from COULOMB_POTENTIAL_RATIO_ARRAY to prevent division by zero issues that were causing NaNs in some numpy packages.

Enhancements:
- Update expected results in test cases to reflect changes in COULOMB_POTENTIAL_RATIO_ARRAY.

Chores:
- Bump the version number from 0.0.18 to 0.0.20 in the project.

### DIFF
--- a/particula/__init__.py
+++ b/particula/__init__.py
@@ -20,7 +20,7 @@ from particula.logger_setup import setup
 # u is the unit registry name.
 u = UnitRegistry(force_ndarray=True)
 
-__version__ = "0.0.18"
+__version__ = "0.0.20"
 
 # setup the logger
 logger = setup()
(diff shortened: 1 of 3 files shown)


Gorkowski (approved):
That makes sense. Maybe we should consider raising an error then when 0 ratio occurs or something like that, so we know where a problem starts before it is propagated.

PR #561, opened 2024-12-23
SemiManual Common Units Conversions

Gorkowski:
An approach for common units conversions. 

- [ ] simple conversions like "m" to "km"
- [ ] fix non unique parsing problem, i.e. "mm" vs "m" problem.
- [ ] due more complex conversions like kg/m3 to ug/cm3.

### DIFF
--- a/particula/util/converting/tests/units_test.py
+++ b/particula/util/converting/tests/units_test.py
@@ -1,39 +1,59 @@
 """
 Tests for Converting Units
 """
-
 import unittest
+from particula.util.converting.units import (
+    UnitConverter,
+)  # Assuming the class is saved in unit_converter.py
 
-from particula.util.converting import units
 
+class TestUnitConverter(unittest.TestCase):
+    def setUp(self):
+        """Set up a UnitConverter instance for testing."""
+        self.converter = UnitConverter()
 
-class TestUnitConversions(unittest.TestCase):
+    def test_si_prefix_conversion(self):
+        """Test conversions using SI prefixes."""
+        self.assertAlmostEqual(
+            self.converter.convert("kiloPascal", "Pa", 1), 1000
+        )
+        self.assertAlmostEqual(self.converter.convert("megaWatt", "W", 1), 1e6)
+        self.assertAlmostEqual(
+            self.converter.convert("milliAmpere", "A", 1), 1e-3
+        )
 
-    def setUp(self) -> None:
-        self.conversions = [
-            ("Pa", "bar", 1e-5),
-            ("bar", "Pa", 1e5),
-            ("Pa", "Atm", 1 / 101325),
-            ("Atm", "Pa", 101325),
-            ("degC", "K", 273.15),
-            ("K", "degC", -273.15),
-            ("degF", "K", (32 - 32) * 5/9 + 273.15),
-            ("K", "degF", (273.15 - 273.15) * 9/5 + 32),
-            ("feet", "m", 0.3048),
-            ("m", "feet", 1 / 0.3048),
-        ]
+    def test_base_unit_conversion(self):
+        """Test conversions between base units."""
+        self.assertAlmostEqual(self.converter.convert("meter", "meter", 1), 1)
+        self.assertAlmostEqual(self.converter.convert("liter", "liter", 1), 1)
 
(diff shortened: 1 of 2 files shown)


Gorkowski:
@mahf708 thoughts on this approach?

I think it should be switched to full str matches instead of regex but generally should work.

Then a separate method could be used to parse divides and multiplication

Gorkowski:
Ha, as a user-dev I'm going to add the length and concentration conversion...I don't want to do them each time and I always forget the volume factors

Gorkowski:
We might not be on the same page regarding the extent to which units should be used.

My view is that units should be implemented only within the builder classes, and even then, only within the builder mixins.

The intent is that the builders provide guardrails and safety features, such as clear error messages and convenient input handling. This makes them an ideal onboarding path for new students.

For users who wish to directly call functions or kernels (as in the your example), they can handle the units themselves.

particula/builder_mixin.py
```diff
@@ -39,7 +40,8 @@ def set_density(
         if density_units == "kg/m^3":
             self.density = density
             return self
-        raise ValueError("Density units must be in kg/m^3")
+        self.density = density * convert_units(density_units, "kg/m^3")
```
Gorkowski, line 43: So, are you thinking don't do this?
Which is fine by me. 

The units input option seems kind a dumb, but maybe should stay for new users.

particula/util/converting/tests/units_test.py
Gorkowski, line 1: we don't need this then

particula/util/converting/units.py
Gorkowski, line 1: So, you don't want this util at all in the code, but fine with it in the docs?

I think put it in the code and the docs. If it ends up breaking, then trash it in both locations.

Gorkowski:
> That's fine. So here's my opinion, we have to choose one of these two routes:
> 
> * particula fully support units (re-integrate pint, etc.) and let users run wild with units
> * particula only supports base units, and so it doesn't do any conversion and error handling
> 

Yes, I support option 2. 

The only point that I think we differ is providing a convert_unit example function in the particula utils
My thought is it works and does no harm as long as we don't integrate it anywhere else.
It seems like your saying, just give them that same example function in the docs and them handle the copy/paste?

I made some comments in this code and change it to use pint if installed. Can you check.

Gorkowski:
What if in these setters/mixins we force the unit as a input?

current version
```
def set_density(
        self,
        density: Union[float, NDArray[np.float64]],
        density_units: Optional[str] = "kg/m^3",
    ):
""" docs""""
````

Force units as input
```
def set_density(
        self,
        density: Union[float, NDArray[np.float64]],
        density_units: str
    ):
""" docs""""
```

File: particula/util/converting/units.py [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+    # 1e-9
+    ```
+    """
+    if UNIT_REGISTRY is None:
+        raise ImportError(
+            "Install pint to use unit conversion features: pip install pint"
+        )
+
+    offset_units = ["degC", "degF", "degR", "degK"]
+    if old in offset_units or value is not None:
```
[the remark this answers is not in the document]
Gorkowski: not needed

particula/tests/builder_mixin_test.py
```diff
@@ -23,27 +23,21 @@ def test_density_mixin():
...
     assert "Argument 'density' must be positive." in str(excinfo.value)
 
     # test positive density
-    builder_mixin.set_density(1)
+    builder_mixin.set_density(1, "kg/m^3")
     assert builder_mixin.density == 1
 
     # test array of densities
-    builder_mixin.set_density(np.array([1, 2, 3]))
+    builder_mixin.set_density(np.array([1, 2, 3]), "kg/m^3")
```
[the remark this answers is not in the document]
Gorkowski: handled by pint when converting or not allowed

Gorkowski:
> I am fine with an optional pint dependency. Let me take a look at how best to handle that without messing stuff up. I think I would rather force users to provide unit-ful inputs very explicitly (let me assess)

I modified the density setter, to force units on inputs. I think this makes it clearer.

The pint conversion is optional. 

We can expand to other setters inputs after this PR.

PR #645, opened 2025-02-28
Ion coagulation notebook, update

### LINKED ISSUE #639: Rewrite/UPDATE ion coagulation example
Update the Ion coagulation example, to current particula version.

Gorkowski:
Updated the Ion coagulation notebook, using the new and flatter API function calls.

Also address 4 small sourcery comments previously suggested.

Fixes #639 

## Summary by Sourcery

This PR fixes a few bugs related to radius calculations that could cause the code to crash. It also updates the API function calls.

Bug Fixes:
- Fixes a bug where the kernel radius could be zero, causing a crash.
- Fixes a bug where the particle radius could be non-finite, causing a crash.

## Summary by Sourcery

Fixes potential crashes due to zero or non-finite particle radii during kernel calculations.

Bug Fixes:
- Fixes a bug where the kernel radius could be zero, causing a crash.
- Fixes a bug where the particle radius could be non-finite, causing a crash.


Gorkowski:
> The notebook makes it sound like the hard-sphere kernel shouldn't have any charge-related effects, yet it changes as the charge count increases. What gives?

I always forget that hard-sphere is the idealized electrostatics. I change the text.

PR #690, opened 2025-04-11
Minor edits to readme and chamber simulation

wkchuang:
I changed the final plot to use the same colors for initial and final, fixed a typo in a variable.
Added conda in the readme as an option, I can remove if not desired.


Gorkowski (approved):
nice plot update!

PR #720, opened 2025-04-19
OpenAI theory docs

### LINKED ISSUE #708: Add prompt context and OpenAI usage
In the Theory section, add a page on the different models and their best applications. Also add this to the SI for the paper.

Gorkowski:
Explanation of OpenAI models and how they get use with Particula documentation.

Fixes #708


Gorkowski:
> Also, let's ensure we have something to back these docs? As in, where did we get this info from?

we'll just cite...Personal correspondence with OpenAI o3  
🤣

Gorkowski:
Made a separate issue to fix the side bar and docs linking #722

PR #831, opened 2025-12-03
update restrict

Gorkowski:
add opencode tool for maintainers


Gorkowski:
/oc 
does the opencode tool work

